PR #2, opened 2017-01-20
Adding package.json file

sherakama:
# Ready for review

* For NPM integration

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,25 @@
+{
+  "bugs": {
+    "url": "https://github.com/su-sws/decanter/issues"
+  },
+  "description": "A collection of SCSS patterns for theme development.",
+  "homepage": "https://github.com/su-sws/decanter",
+  "keywords": [
+    "css",
+    "mixins",
+    "sass",
+    "scss"
+  ],
+  "license": "GPLV2",
+  "name": "decanter",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "decanter",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/su-sws/decanter.git"
+  },
+  "version": "0.1.0",
+  "dependencies": {
+    "bourbon": "~5.0.0-beta.7",
+    "bourbon-neat": "^2.0.0-beta.2",
+    "font-awesome": "^4.6.3"
```
sherakama: I can go both ways on adding this or not.

kgcreative: These should be `devDependencies`, rather than `dependencies`, since there is no runtime. I believe this will fetch those projects into `node-modules`, which may end up throwing conflicts if the root project uses different versions. We should test this.

sherakama: I think we do want the version conflicts. If we are using bourbon 2.0 and someone is using the ~1.0 version in their app then they won't be able to compile or they will have errors. I think any version conflict from declaring these as dependencies is a helpful error.

kgcreative: That works for me!

PR #3, opened 2017-01-22
update package.json

kgcreative:
- clarify devDependencies vs dependencies
 - clarify specific bourbon & bourbon neat versions
 - clarify ~4.7.0 font-awesome version

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "url": "https://github.com/su-sws/decanter.git"
   },
   "version": "0.1.0",
-  "devDependencies": {
-    "bourbon": "5.0.0-beta.7",
-    "bourbon-neat": "2.0.0-beta.2",
+  "dependencies": {
+    "bourbon": "^5.0.0-beta.7",
+    "bourbon-neat": "^2.0.0-beta.2",
     "font-awesome": "~4.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "url": "https://github.com/su-sws/decanter.git"
   },
   "version": "0.1.0",
-  "dependencies": {
-    "bourbon": "~5.0.0-beta.7",
-    "bourbon-neat": "^2.0.0-beta.2",
-    "font-awesome": "^4.6.3"
+  "devDependencies": {
+    "bourbon": "5.0.0-beta.7",
+    "bourbon-neat": "2.0.0-beta.2",
+    "font-awesome": "~4.7.0"
   }
 }


package.json
```diff
@@ -17,9 +17,9 @@
     "url": "https://github.com/su-sws/decanter.git"
   },
   "version": "0.1.0",
-  "dependencies": {
-    "bourbon": "~5.0.0-beta.7",
-    "bourbon-neat": "^2.0.0-beta.2",
-    "font-awesome": "^4.6.3"
+  "devDependencies": {
+    "bourbon": "5.0.0-beta.7",
+    "bourbon-neat": "2.0.0-beta.2",
```
kgcreative: i'm specifying 5.0.0-beta.7 and 2.0.0-beta.2 without any `~` `>` `^` `=` modifiers since we want to be very deliberate about what might get pulled in. Font awesome I want to auto update, but it will still be good to come back and update as the package continues to evolve.

sherakama: Re Dev vs Dependency:

From a stackoverflow answer:
```
dependencies are installed on both:
npm install from a directory that contains package.json
npm install $package on any other directory

devDependencies are:
also installed on npm install on a directory that contains package.json, unless you pass the --production flag
not installed on npm install "$package" on any other directory, unless you give it the --dev option.
are not installed transitively.
```

So basically, dependencies are items that have to be downloaded for the app to run or function and devDependencies are there to develop on it. Since this is not really an "app" I think it is fair to include anything we use in the library of files as a dependency.

In my humble opinion, semantic versioning and the special modifiers are the greatest strength of a package manager. They allow us developers to automatically get "safe" versions downloaded and applied to our work. The two characters I used were:

```
~version "Approximately equivalent to version"
^version "Compatible with version"
```

These are completely safe to use as they will not jump up a major version and only allow safe bug fix updates to propagate to our work. When 2.0.0 is released I don't wan't to have to go back in to my project to update that as I know it was a safe update. When 2.1.0 is released we will have to update the version again after we validate that it works with our stuff.

package.json
```diff
@@ -17,9 +17,9 @@
...
+    "font-awesome": "~4.7.0"
```
sherakama: Yay, newer font awesome!

package.json
kgcreative: That makes sense to me and I think it's a safe path moving forward. Thoughtbot tends to be (sometimes a little too) conservative with their releases, and so far I haven't run into any breaking changes that haven't been announced.

PR #4, opened 2017-01-23
Initialize bitters scaffold

kgcreative:
#READY
Tons of changes. See latest commit. We should merge to master and continue from there.
---
Scaffold bitters into 7-1 architecture
 
 • master.scss -- with caveats. Many discussion points here 
 • base/_base.scss -- most everything from base is less framework and more opinionated extensions that extend normalize.css. These are all theme configurable, so they would inherit _variables.scss settings from both decanter and/or the base theme.
 • base/_buttons.scss
 • base/_fonts.scss
 • base/_forms.scss
 • base/_lists.scss
 • base/_media-elements.scss
 • base/_tables.scss
 • base/_typography.scss
 • layout/_layout.scss -- with caveats. Discussion point here.
 • utilities/_color.scss -- color picker function. Custom code written by @kgcreative
 • utilities/_constants.scss -- some discussion points here.
 • utilities/_grid-settings.scss -- discussion points re: neat grid setups
 • utilities/_variables.scss -- some discussion points re: naming conventions needed
 • vendors/_unit-conversion.scss -- discussion points here. It's technically a core function, but it's not maintained by us.
• vendors/README.md
• vendors/_unit-conversion-README.md


kgcreative:
Still lots of integration, comment, and documentation work to be done in this branch, but I wanted to kick off with something to be able to start having productive discussions around. Some testing will be needed here, so i'll probably import this into a blank brunch project for some testing, as well as our WP test instance.

base/_base.scss
```diff
@@ -0,0 +1,28 @@
+@charset "UTF-8";
```
sherakama: Code climate is giving an error with the message "Prefer single quoted strings". I think that is a little too nit picky. Would you object to disabling that error check?

base/_buttons.scss
```diff
@@ -0,0 +1,63 @@
...
+// see http://bitters.bourbon.io
+// dependent on utilities/color.scss
+// -----------------------------------------------------------------------------
+
+
+// Variables in use on this file.
+// Variables are defined with !default flags, and can be overwritten in
+// `utilities/_variables.scss`
+// -----------------------------------------------------------------------------
+$action-color: color(generic, blue) !default;
```
sherakama: Code climate is giving the error: "Color should be written in hexadecimal form"
I think we can disable that check. Any objections?

base/_base.scss
sherakama: I think this file name should be reserved for the "roll up" partial include file. I propose renaming this file to _box-model.scss and creating a base.scss for all of the imports.

base/_buttons.scss
```diff
@@ -0,0 +1,63 @@
+@charset "UTF-8";
+
+// buttons.scss
+// This file contains all styles related to the button element.
+// see http://bitters.bourbon.io
+// dependent on utilities/color.scss
```
sherakama: If this file is dependant on another file for a mixin, extend, or function placing an @import makes sense to me. I like the top level comment heading _dependencies_. We should strive hard to not @import files which generate markup but rather only have mixins and functions.

base/_buttons.scss
```diff
@@ -0,0 +1,63 @@
...
+// -----------------------------------------------------------------------------
+
+
+// Variables in use on this file.
+// Variables are defined with !default flags, and can be overwritten in
+// `utilities/_variables.scss`
+// -----------------------------------------------------------------------------
+$action-color: color(generic, blue) !default;
+$reverse-font-color: color(generic, white) !default;
```
sherakama: Each of these will need inline comments. I am unsure what reverse-font-color is or why I would need it.

base/_buttons.scss
```diff
@@ -0,0 +1,63 @@
...
+$base-font-size: 1em !default;
+$base-line-height: 1.5 !default;
+$base-spacing: $base-line-height * 1em !default;
+$small-spacing: $base-spacing / 2 !default;
+$base-duration: 150ms !default;
+$base-timing: ease !default;
+
+// All buttons styling
+// -----------------------------------------------------------------------------
+#{$all-buttons} {
```
sherakama: The properties and choices here look good to me.

base/_fonts.scss
```diff
@@ -0,0 +1,39 @@
...
+// Full font families:
+// $google-source-code: 'https://fonts.googleapis.com/css?family=Source+Code+Pro';
+// $google-source-serif: 'https://fonts.googleapis.com/css?family=Source+Serif+Pro:400,600,700';
+// $google-source-sans: 'https://fonts.googleapis.com/css?family=Source+Sans+Pro:200,200i,300,300i,400,400i,600,600i,700,700i,900,900i';
+// $google-noto-sans: 'https://fonts.googleapis.com/css?family=Noto+Sans:400,400i,700,700i';
+// $google-noto-serif: 'https://fonts.googleapis.com/css?family=Noto+Serif:400,400i,700,700i';
+// Conditional Declarations
+// -------------------------------------------------------
+
+@if(global-variable-exists(source-code)) {
```
sherakama: Do these need to be conditional? Is there any harm in declaring them in this file?

base/_fonts.scss
```diff
@@ -0,0 +1,39 @@
+// -----------------------------------------------------------------------------
```
sherakama: I like the sass docs format for file annotations. We should follow it for all of our files:
http://sassdoc.com/file-level-annotations/

base/_base.scss
kgcreative: Yeah, I'm ok with that. Some instances ca for " ' ' " nested quotes, and it's so conditional that I'm ok killing that check

kgcreative: I'm on board with creating roll ups for partials. In this case, _box-model makes sense here. I was struggling to figure out where to drop these in

base/_buttons.scss
kgcreative: Bitters generates a bit of markup, in addition to the needed variables, mixins, functions and placeholders. Most of that markup is concerned with provided sensible typography and base HTML styling, in addition to setting good patterns for later component builds. Maybe this can all be included conditionally in rollups. Let's discuss this further.

base/_buttons.scss
kgcreative: Disable away.

base/_fonts.scss
kgcreative: Yeah, I need to go through and squash a bunch of these. It's a tedious process, but well with it

base/_fonts.scss
kgcreative: Yes, these add a lot of weight to the downloaded files. We want to be picky about which fonts we're importing. I haven't quite settled on a sensible minimal font loading strategy here, so there's room for improvement

base/_base.scss
kgcreative: Ugh autocorrect on the train is not my friend

base/_buttons.scss
kgcreative: Scss has if function/mixin exists calls. We can wrap imports in these so if they were already called we don't double them.

sherakama: How about adding import-once as a dependency: https://github.com/Compass/compass/blob/master/import-once/README.md

That would save us from having to write conditional statement and would make the @import statement function more like I would expect it to. With import-once we should be able to declare the dependencies using import statement on each sass file without extra overhead or markup.

base/_base.scss
sherakama: Removed

base/_buttons.scss
sherakama: This one: https://www.npmjs.com/package/sass-import-once

base/_buttons.scss
sherakama: disabled.

base/_fonts.scss
sherakama: I can help with that as well.

base/_fonts.scss
sherakama: I wonder if it would make sense to separate these in to separate files and create a function/mixin that @imports the correct files. I did a quick google search for font loading patterns and I did not get much back. I'll add an issue to track this idea for a later version.

base/_base.scss
sherakama: 👍

base/_buttons.scss
kgcreative: I'd be willing to test it out. It seems like it will definitely help and make our package more robust. 👍

base/_fonts.scss
kgcreative: 👍 - it would be good to be able to conditionally pull in only the necessary weights. I bet you we can write a mixin that does this.

base/_fonts.scss
kgcreative: 👍🏼

sherakama:
One thing I am unsure about is how normalize fits in with this. Do we add it as a dependency, as documentation, inline, or something else?

kgcreative:
RE: [Normalize](https://github.com/SU-SWS/decanter/pull/4#issuecomment-275747705):

This is an opinionated decision. I have in the past:

- Included it in <head> right before global CSS
- Included it as the first @import in master.scss

Normalize has an NPM module, so it makes updating it easy. My current personal preference would be to include it as a dependency, add it to sass import paths. In my case, I generally @import it after all the configurations and helpers, before font awesome, and before base, but i'm not married to that idea. 

Bitters mentions in their [github project](https://github.com/thoughtbot/bitters) that

> Bitters is made to work alongside a CSS reset; not replace one. Our suggested reset is Normalize.

We could follow the same pattern and simply suggest Normalize as a good default, and leave implementation up to each developer.

PR #11, opened 2017-02-15
Compile

sherakama:
# Not Ready

But opening in order to start conversation.


kgcreative:
Do we want to add a `core/` folder to be the scss root for this project? This would allow us to have the readme files, potentially sassdoc, etc in other folders and would allow us to configure import paths for sass to grab npm/decanter/core

sherakama:
I was playing around with sassdoc and moving the scss files to a folder such as core make it easier for the compiler. 

I chose the name 'src' for now but core works for me as well.

kgcreative:
`core` is the convention both bourbon and neat use. In the case of this being an importable module, it makes sense

sherakama:
Changed to core.

sherakama:
While we are moving things around. I've seen the convention of removing the leading "_" for the roll up file names. eg: /core/utilities/settings/settings.scss

I like this as it separates out the file from the other partials. Do you have any objections to this?

kgcreative:
I'm ok with this. I was achieving something similar by having rollups be one level down. By default, sass doesn't compile _underscore files, while it tries to create compiled CSS for non underscore files, so we'd need to test the behavior.

kgcreative:
Regarding the `base/` folder - do you think it would make more sense to rename it `boilerplate`? I don't have a strong opinion on this one, but it might be more semantically specific in terms of what those files actually are

sherakama:
I've got some thinking to do about that particular folder. As it does have markup it should only ever be included once in a project but because it uses mixins, extends, and variables from the rest of the project it is going to be difficult to ensure that the includes load the vars, bourbon, and neat in the correct order. If I don't import the requirements then sassdoc yells at me and renders incomplete documentation.

sherakama:
I would like to have the 'used by' list be a complete list so in the future we will know the impact of our changes. 

screen shot 2017-02-15 at 2 38 16 pm

kgcreative:
In my latest refactor, if you include 'settings' - it loads global settings, bourbon, neat, and variables in the correct order. Settings should be included first before anything else.

kgcreative:
Merging for now so we can have some of the file system changes in the master branch. I think this is a good start. Let's test out un _underscored rollups, but at this stage they should be ok.

PR #12, opened 2017-02-16
Integration

kgcreative:
# NOT READY

Integrating this into a brunch build, and learning what changes we need to make in order for bourbon, neat, boilerplate and other settings to be instantiated and work.

For example, I had to split the master rollup into two:
`decanter-preload and decanter-core`
 - `global-settings` in your project should be loaded before `decanter-preload`
 - then your `variables` file
 - then `decanter-core`
 - then the rest of your project files

Because decanter is so dependent on bourbon and neat functions, we should go ahead and load them as part of `decanter-preload` so they are called at the right moment and inherit the correct local variables and maps.

I also needed to move charset definitions above the core comments. Otherwise compilation failed.
For now I removed loading settings on every individual module, i'll have to test it again, but i was getting an infinite compile loop and didn't have time to look into it.

I'll continue to evolve this branch, for now, my package.json calls
`"decanter": "github:SU-SWS/decanter.git#a52e672d460328ae137e4eefee75b996a65c7944",`
where the hash is the commit hash of whatever the latest fixup to this PR is.


sherakama:
Sorry about the charset thing. I was working through bugs in sassdoc and thought that the file level comment had to be the first thing in the document. Turns out not.

kgcreative:
Todo: We need to talk about `base`
Specifically, we need to decide if:
- We rename `base` to `boilerplate` and provide a mechanism for importing it conditionally
or
- We provide configuration options for base elements to be able to be modified from `variables`
or
- We remove base all together and we provide it as an optional boilerplate that people can use for scaffolding their own projects.

PR #13, opened 2017-02-20
Integration2

sherakama:


kgcreative:
Hmm, I'm liking the idea of removing the settings files and either i standing them in the partials, or in default maps that can be merged. Definitely a topic for discussion tomorrow

sherakama:
Yeah, lots to talk about.

I took a stab at integrating today and the many different ways to import what was needed was throwing me off so I started simplifying and kind of went nuts. Breaking out the variables in to smaller files really helped troubleshoot the order of operations I was looking for. I was finally able to break it out an import just the base styles where I wanted to. 

I also threw in an example landing page which uses a full compile of decanter. 

So those two scenarios are working so far in this branch. The next steps for me were to create a couple of responsive layouts and a component. 

Anyhow, this gives us a lot to chat/work on for our session tomorrow.

kgcreative:
I finally also grokked neat's instance of grid-media(), so we can kill the @media mixin in favor of that. I need to spend some time showing you how centered container works as well, and possibly extending it to account for your specific use case, which is containers that are a fixed instead of fluid width inside breakpoints other than the widest one.

Issue #15, opened 2017-02-24
namespace rollup files


kgcreative:
During integration, sometimes it's hard to know whether you are pulling a file from your own library, or from Decanter's. Refactor rollups to include decanter namespace.

sherakama:
hmm. interesting. 

Perhaps when we add our includes paths we only add 'node_modules' and not the full path to decanter. So to get everything your import would be:

```
@import 'decanter/decanter';
```

and to get at portions it would be

```
@import 'decanter/core/utilities/utilities';
```

sherakama:
This would be versus what we have when our include paths are set to 'node_modules/decanter'

eg:
```
@import 'decanter';
@import 'core/utilities/utilities';
```

kgcreative:
I thought about that, but that has other implications. This allows us to namespace rollups, without having to remove underscores. So any '_decanter-*' file we know is a rollup, and doesn't contain actual funtionality, plus, when importing, this makes it clearer when you are calling a decanter-specific rollup versus your own.

I'm ok with actual mixins, base files, functions, etc not being namespaced, but namespacing the rollup files gives us much more flexibility when developing.

kgcreative:
(there's something that feels dangerous about adding npm path to our scss import paths, vs adding individual components, but I haven't quite figured out what.)

sherakama:
From what I have seen a number of projects are using this: http://eyeglass.rocks/

I haven't fully grokked it yet but this seems to be a standard or convention for managing node modules in sass.

kgcreative:
I believe both bourbon and neat can be eyeglass modules. I haven't quite grokked it yet, so i've been kicking that can down the road, since I don't want to add too much weight to our libraries

kgcreative:
(this can be closed)

PR #16, opened 2017-02-24
Namespace rollup files:

kgcreative:
# READY

 - base -> _decanter-base
 - variables -> _decanter-base-variables
 - variables -> _decanter-config
 - vendors -> _decanter-vendors
 - components -> _decanter-components
 - utilities -> _decanter-utilities
 - themes -> _decanter-themes


kgcreative:
Addresses #15 
This makes integration much easier, and allows us to pull in rollups that we know belong to decanter. I also split variables into decanter-config, and decanter-base-variables, the latter of which is pulled in the base rollup. This allows us to use the decanter functions to specify variables.

In my integration use cases, i'm finding that I want to import variables at different points.

So, I can call:

```
@import 'my-global-settings'; // which has variables that override bourbon, neat, and decanter-config

@import 'decanter-vendors';
@import 'decanter-utilities';

@import 'my-variables'; // which can override base variables

@import 'decanter-base';
@import 'decanter-layout';
@import 'decanter-themes';
```

I actually don't know that 'decanter-themes' and 'decanter-layout' will actually need to exist. Those can either belong in base, or are either functions or mixins.

PR #20, opened 2017-02-25
standards and enforcement

sherakama:
# Some cleanup and a couple of changes.

* Rename decanter-base.scss to _decanter-base.scss
* When you remove the global defaults from the color mixin it turns in to just a pass through function. Can we remove it or change it to be useful again.
* Removed neat-omega from the decanter-mixins roll up as it is not a part of this project. neat-omega should be imported where it is used only.
* Added flex-container mixin to roll up
* Removed all global variables from mixins and functions.


kgcreative:
I'm not opposed to renaming `_retrieve-color` to `color` and removing the pass-through function. The color function does need to call a default map, so it is dependent on it existing, but maybe the default map can be configured to accept a setting

kgcreative:
Flex container will use the neat-omega mixin. It's not a decanter variable, but it does depend on neat. Maybe that can be moved to after we import neat

core/utilities/variables/_decanter-variables.scss
```diff
@@ -15,8 +15,8 @@
 // Neat with decanter custom settings.
 @import
   'variables-neat',
-  'neat';
-
+  'neat',
+  'omega';
```
kgcreative: omega is a custom neat module.

kgcreative:
This is ready to integrate if you you're good with it

PR #21, opened 2017-02-25
Twigs

sherakama:
# READY

Layouts galore!


kgcreative:
Get out of my head! I started refactoring centered section because the first proto Mixin  was done before I fully grokked grid media

kgcreative:
`.column-second.collapse`

You don't need the extra wrapper. You can add a grid collapse to any grid object.

kgcreative:
Also, for a variable height grid, or for a centered flexbox grid, you will end up needing `@include omega(4n);` on the parent container to ensure the children properly clear their grids.  You don't need omega if it's a flex grid that is not centered.

sherakama:
If I follow your example markup

```
section {
  @include centered-container($full-container);
  .grid-object {
    @include grid-container;
    @include grid-collapse; // (if I want it to align to the edge)
   .six-cols {
       @include grid-column(1 of 2);
       @include grid-collapse; //so it can accept nested objects)
         .three-cols {
             @include grid-column(1 of 2);
         }
      }
    }
}
```

The HTML for that would be 

```
        <section>
          <div class="grid-object">
            <div class="six-cols">
              <div class="three-cols">
                content.
              </div>
            </div>
          </div>
        </section>
```

That is four nested html elements. I was hoping for three. I currently have:

```
<div class="container">
  <div class="grid-column(4)">
    <div class="grid-column(2)">
   content.
    </div>
  </div>
</div>
```

Without that fourth if I try to collapse any of those divs I get wacky results. I'll revise the html.

sherakama:
screen shot 2017-02-28 at 10 35 03 pm

I'm not getting the alignment that I would like to see even with the appropriate collapse options. I suspect it has something to do with the fact I am grid-shift() ing the sidebar in to place.

kgcreative:
The impetus for me to create the centered-container mixin is so we could get rid of grid object. The only reason you would need it is if you have other html content interspersed in the same section

if you don't, you can just
HTML:
```
<section>
  <div class="one-half nested">
    <div class="one-half">
    content.
    </div>
    <div class="one-half">
    content.
    </div>
  <div class="one-half">
    content.
  </div>
</section>
```

scss
```
section {
@include centered-container(12, $grid-collapse: true);
  .one-half {
    @include grid-column(1 of 2);
  }
  .nested { @include grid-collapse; }
}
```

kgcreative:
grid-shift() uses relative positioning to shift a grid item without changing the space it occupies.
grid-push() uses margin adjustments instead.

instead of grid shifting both columns, try grid-pushing the content to the right.

examples/layout/three-column/sunset-delorean-right/index.html
```diff
@@ -11,45 +11,40 @@
 
         <!-- the main content //-->
         <div class="column--first">
-          <div class="collapse">
-            <div class="column--first-header">
-              <h2>Page title that spans two columns</h2>
-            </div>
+          <h2>Page title that spans two columns</h2>
 
+          <div class="nested-grid">
```
sherakama: @kgcreative, thanks for this example. This does exactly what I want it too but it still has one extra div just for the collapse. This is a better place for it than where I was placing it.

core/utilities/variables/_containers.scss
```diff
@@ -2,10 +2,18 @@
...
-    xl:     ( columns: 12, grid-media: $media-xl,       v-space: 1em, max-width: true,  grid-collapse: false ),
-    print:  ( columns: 12, grid-media: $media-print,    v-space: 1em, max-width: false, grid-collapse: false ),
+  xs: ( columns: 12, grid-media: $media-xs,       v-space: 1em, max-width: false, grid-collapse: false ),
+  lg: ( columns: 12, grid-media: $media-lg-only,  v-space: 1em, max-width: false, grid-collapse: false ),
+  xl: ( columns: 12, grid-media: $media-xl,       v-space: 1em, max-width: true,  grid-collapse: false ),
+  print: ( columns: 12, grid-media: $media-print, v-space: 1em, max-width: false, grid-collapse: false ),
+) !default;
+
+/// Default Responsive Container Collapsed
+$default-container-collapsed: (
```
sherakama: @kgcreative: Should we keep?

sherakama:
@kgcreative,

I think that is everything I wanted to create as far as layouts go.
Now for a few components.

Please review and squash and merge if appropriate.

PR #23, opened 2017-02-27
Refactor container system

kgcreative:
# READY
 - rename $max-width to $max-container-width
 - move $gutter from variables-neat to variables-decanter
 - move $grid-columns from variables-neat to variables-decanter
 - move $max-content-width from variables-neat to variables-decanter
 - move $max-container-width from variables-neat to variables-decanter
 - remove centered-section proto-code
 - refactor centered-section to centered-container
 - refactor _set-container-padding function to reflect internal use, and to accept max columns from the passed grid
 - refactor _set-container-padding function to accept a pixel value for max width, or to default to the project’s $max-container-width otherwise.
 - create new responsive-container mixin, which accepts a map.

I'm debating whether `centered-container` and `responsive-container` should exist in `components` or in `mixins` - For now, i put `centered-container` in mixins, and `responsive-container` in components, but i'm open to moving them around as you think it's more useful


Gruntfile.js
```diff
@@ -36,7 +36,7 @@ module.exports = function(grunt) {
         sourceMap: true,
         // This controls the compiled css and can be changed to nested, compact or compressed.
         outputStyle: 'expanded',
-        precision: 5
+        precision: 10
```
kgcreative: Technically bootstraps need 8, but it doesn't hurt to go to 10. most browsers ignore past a certain threshold anyway.

core/utilities/functions/_set-container-padding.scss
```diff
@@ -1,40 +1,35 @@
...
-  $_max-width: false
+@function _set-container-padding(
+  $columns: null,
+  $grid: $neat-grid,
+  $v-space: 1em,
+  $max-width: false
 )
 {
-  $_gutter: _retrieve-neat-setting($_grid, gutter);
+  $columns: _neat-column-default($grid, $columns); // Retrieves max columns from specified grid if empty, or the specified number of columns otherwise
```
kgcreative: This is a pattern I picked up from the neat library. Basically all this does is if the number if columns is null, it grabs the number of max column from the grid and defaults to that. It's a shortcut to creating a full width container. Technically, if you pass it a number, it returns the same number you passed it, hence why it keeps it's same name.

kgcreative: Although in this instance, I would be ok using $_columns for this. At least it would be obvious that we're using a different value.

sherakama:
Thanks!

sherakama:
You are very speedy.

kgcreative:
I happened to be working on this part of the code on homesite. It worked out! :D

core/utilities/functions/_set-container-padding.scss
```diff
@@ -1,40 +1,35 @@
...
+// Can be constrained to a specific max width, or to the project's global `$max-container-width`.
 
-@function set-container-padding(
-  $_cols: null,
-  $_v-space: 1em,
-  $_grid: null,
-  $_max-width: false
+@function _set-container-padding(
+  $columns: null,
+  $grid: $neat-grid,
```
kgcreative: Yeah, this one is highly dependent on neat, which is why i'm comfortable with making that assumption

PR #28, opened 2017-03-08
Items per row

kgcreative:
# READY
With Examples!


sherakama:
Nice. These are great. 

I've added the two mixins to the main index.html directory file at /examples/index.html

PR #33, opened 2017-03-15
Conditionalize BPH helper output

sherakama:
Wraps the whole output of the id in a conditional in order to avoid markup bleeding

I'm getting this css outputted multiple times in my integrations as anytime I include the components roll up it gets put in. Can we go one step further and make this a mixin instead of a hard ID output?

I'm happy with just the conditional for now.

### DIFF
--- a/core/base/_decanter-base.scss
+++ b/core/base/_decanter-base.scss
@@ -13,12 +13,12 @@
 ////
 
 @import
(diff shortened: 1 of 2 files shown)


sherakama:
Woot thanks. I'll add an issue.

Issue #34, opened 2017-03-15
Breakpoint helper improvements


sherakama:
- Breakpoint Helper should be changed to a mixin
- It should consume a map
- The mixin iterates over the map (a la responsive container) and outputs all the breakpoints and colors as defined in the map.

PR #36, opened 2017-03-16
Breakpoint helper improvements

kgcreative:
Addresses #34
- Add ‘helpers’ to base. Instantiate ‘breakpoint helper’ there. This is where things like ‘.center’ ‘.sr-only-text’, etc and other helper classes can exist if we decide we want them. Maybe helpers can have true/false configuration settings somewhere.


sherakama:
I like this separation. 

My only question is does this helper belong in 'components' instead of base? I think this could live as it's own thing there.

kgcreative:
I don't have a strong feeling of where it belongs. I figured a 'helpers' rollup in base could conditionally call the component, in addition to things like `.center` `.sr-only-text`, `.sr-only-element`, any visibility classes, maybe some grid classes if we ever wanted to go that route. I'm also ok with not calling that component anywhere and removing the true/false flag, and just let individual developers instance those as needed.

sherakama:
In our very own comments

> Base rules consist of styling for HTML elements only, such as used in a CSS
> reset or Normalize.css. Base rules should never include class selectors.

We should play by there rules.

If you don't have a strong feeling about this I would like to move it to components.

kgcreative:
- Refactored helpers.
- Introduced some placeholders

kgcreative:
@sherakama - bump

sherakama:
Ah sorry, thanks for the reminder. I'll look at this later today.

sherakama:
@kgcreative,

I put the components in their own folder as this directory will likely explode next week with new components and we want to keep them separate. Helpers became a component as well. Not sure if that is the long term place for that kind of thing yet or not. Fixed up a few roll ups and compiled css.

kgcreative:
Works for me! Thank you

PR #37, opened 2017-03-27
Document and add to Color System

kgcreative:
- Add `color-theme` mixin
 - Add `color-themes` mixin
 - Document color system (with examples)
 - update default style for `pre` block
 - Split up layout and theme mixins into folders
 - Re-run grunt-sass


sherakama:
This is pretty darn nifty

PR #42, opened 2017-04-03
Documentation

kgcreative:
# NOT READY
Starting to beef up sassdoc documentation. Also providing some examples.


core/utilities/variables/_typography.scss
```diff
@@ -2,44 +2,83 @@
...
 ////
 
-// Individual Font family definitions
-// ----------------------------------
-
 /// Font Stack Monospace.
+///
 /// This is outputted via `font-family: $font-stack-monospace;`
+/// @type list
+/// @example - SASS
```
kgcreative: this should be @example - SCSS (we are using scss syntax, not sass syntax)

core/utilities/mixins/theme/_color-themes.scss
```diff
@@ -1,5 +1,7 @@
 @charset "UTF-8";
 
+@import '../../variables/colors';
```
sherakama: @kgcreative there is a dependency on a variable in this file so I am adding the import declaration in order get stuff to compile.

PR #46, opened 2017-05-17
Responsive grid collapse.

sherakama:
* Adds a responsive grid collapse mixin to account for mixed grids at breakpoints.
* Adds the grid parameter to the centered container mixin so that it doesn't default to the neat grid and uses the passed in parameter.


kgcreative:
see:

https://github.com/SU-SWS/decanter/blob/master/core/utilities/mixins/layout/_responsive-container.scss

and:

https://github.com/SU-SWS/decanter/blob/master/core/utilities/mixins/layout/_centered-container.scss

The centered container mixin accepts a grid-collapse parameter, so you can configure the responsive container to collapse in the map itself. We shouldn't need a second responsive grid collapse mixin for this.

(When you are inside a grid-media call, the default $neat-grid is overriden with the current grid. This is by design from the folks at Thoughtbot, which is why we don't need a sepparate passed grid parameter)

PR #48, opened 2017-05-17
Prevent font loading leakage.

sherakama:
When I include just the utilities roll up I was getting the base fonts loaded each time.

The base typography file seems like a better home for this function.


kgcreative:
I like this a lot.

PR #54, opened 2017-11-24
Decouple

sherakama:
This is a discussion PR not something to be merged.

Notes placed below but we can go over them in person during a working session.


Gruntfile.js
```diff
@@ -3,12 +3,23 @@ module.exports = function(grunt) {
     pkg: grunt.file.readJSON('package.json'),
     sass: {
       options: {
+        includePaths: [
+          "node_modules/bourbon/app/assets/stylesheets",
```
sherakama: Include paths let us use NPM for lib dependency management. This is how it is done.

Gruntfile.js
```diff
@@ -3,12 +3,23 @@ module.exports = function(grunt) {
...
+          "node_modules",
+          "scss"
+        ],
         lineNumbers: true,
         sourcemap: 'none'
       },
       dist: {
         files: {
           'css/decanter.css': 'scss/decanter.scss',
+          'css/decanter-no-markup.css': 'scss/decanter-no-markup.scss',
```
sherakama: Two new example files. One that includes decanter without any markup so developers can use all the functions, mixins, or start to add just the parts they want.

Gruntfile.js
```diff
@@ -3,12 +3,23 @@ module.exports = function(grunt) {
...
+          'css/decanter-grid-only.css': 'scss/decanter-grid-only.scss'
```
sherakama: Shows an example of how to pull just the grid out of decanter.

Gruntfile.js
```diff
@@ -54,7 +65,6 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-cssmin');
   grunt.loadNpmTasks('grunt-contrib-uglify');
   grunt.loadNpmTasks('grunt-sass');
-  grunt.loadNpmTasks('grunt-contrib-sass');
```
sherakama: Don't need both sass and contrib sass.

package.json
```diff
@@ -16,10 +16,16 @@
   "bugs": {
     "url": "https://github.com/SU-SWS/decanter/issues"
   },
+  "dependencies": {
```
sherakama: External libraries should be managed through a package manager and never committed directly to the project repository.

package.json
```diff
@@ -16,10 +16,16 @@
...
+    "bourbon": "^4.2.7",
```
sherakama: This version is what USA Gov Standards was using but it is behind and has much deprecated code in it. This will need to be updated.

package.json
```diff
@@ -16,10 +16,16 @@
...
+    "bourbon-neat": "^1.8.0",
```
sherakama: This is an obsolete version and completely changes in 2.0. This needs to be updated and the grid-system will need to adjust to accommodate.

package.json
```diff
@@ -16,10 +16,16 @@
...
+    "font-awesome": "~4.7.0",
+    "neat-omega": "^3.0.0",
+    "normalize.css": "^6.0.0"
+  },
   "devDependencies": {
     "grunt": "^1.0.1",
     "grunt-contrib-cssmin": "^2.2.1",
-    "grunt-contrib-sass": "^1.0.0",
```
sherakama: Dont need this and grunt-sass

scss/components/atoms/_button-unstyled.scss
```diff
@@ -0,0 +1,5 @@
+@charset 'UTF-8';
```
sherakama: New atom directory is not needed but might be helpful. Example of an item is an unstyled button.

scss/components/atoms/index.scss
```diff
@@ -0,0 +1,6 @@
+@charset 'UTF-8';
```
sherakama: After reading several documents about partial dependency management it turns out the the industry convention for roll up files is to use an index.scss file. Future versions of sass will automatically include these if not defined in the `@import` statements

scss/components/molecules/_alerts.scss
```diff
@@ -0,0 +1,95 @@
...
+$decanter-custom-alerts: () !default;
+
+$decanter-alerts: (
+  success: rgba(23, 94, 84, 0.6),   // $color-palo-alto at 60% opacity
+  warning: rgba(234, 171, 0, 0.6),  // $color-sun at 60% opacity
+  error:   rgba(177, 4, 14, 0.5),   // $color-bright-red at 50% opacity
+  info:    rgba(0, 152, 219, 0.6),  // $color-sky at 60% opacity
+);
+
+$alerts: map-merge($decanter-alerts, $decanter-custom-alerts);
```
sherakama: $alerts is a global variable that is too generically named. Instead it would be best to set this to $decanter-alerts with this map merge.

scss/components/molecules/_navigation.scss
```diff
@@ -0,0 +1,380 @@
...
+
+///
+/// NAVIGATION COMPONENT
+///
+
+// Navigation variables
+$z-index-nav: 9000;
+
+// Main menu navigation
+@mixin nav-border {
```
sherakama: I should have moved this to it's own mixin file. Ooops.

scss/components/molecules/_navigation.scss
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  li {
+    font-size: $h5-font-size;
+    margin-bottom: 0;
+  }
+}
+
+// Mega menu navigation
+@mixin outer-megamenu {
```
sherakama: This should be it's own mixin file.

scss/components/molecules/_navigation.scss
```diff
@@ -0,0 +1,380 @@
+@charset 'UTF-8';
+
+///
+/// NAVIGATION COMPONENT
```
sherakama: This file is very long and contains three different navigational items. Main / Side / Mega. It would be best to split these up in to their own files.

scss/components/organisms/accordions/index.scss
```diff
@@ -0,0 +1,9 @@
+@charset 'UTF-8';
+
+@import
+  "accordion-button-styles",
```
sherakama: This component had a lot of mixins in it. Are these all really to be re-used in other components?

scss/decanter.scss
```diff
@@ -1,20 +1,39 @@
-@charset "UTF-8";
+@charset 'UTF-8';
 
+///
+// Load Order
```
sherakama: Unfortunately, until the next version of sass comes out we will have to be very careful with the load order of things. There is no (clean/easy/ideal) way to prevent multiple includes and variable dependency management until this new version comes out. 

Bourbon, Neat, Bitters, Refils and USA Gov standards all follow the same convention as listed below.
eg: https://github.com/18F/web-design-standards/blob/develop/src/stylesheets/uswds.scss

In order to ensure that all variables have been set, all mixins are available, and everything is ready for the elements and components we need to have them loaded and ready before we try and load 5. and 6.

Separation is also key for being able to use Decanter without duplicate markup.
Levels 1, 3, & 4 MUST NOT generate any markup when compiled.

scss/decanter.scss
```diff
@@ -1,20 +1,39 @@
...
+  "neat",
+  "font-awesome",
+  "normalize";
+
+// 3. Post Vendor Settings.
 @import
-  'lib/bourbon',
-  'lib/neat',
-  'lib/normalize';
+  "variables/index";
```
sherakama: Variables had to be pulled out as it's own thing as they need to be accessible without bringing along any markup.

scss/decanter.scss
```diff
@@ -1,20 +1,39 @@
...
+// 2. Vendors
+// 3. Post-vendor settings
+// 4. Utilities (functions, mixins, extends)
+// 5. Core Elements
+// 6. Components
 //
 
-// Vendor
+// 1. Pre-vendor settings.
+@import 'variables/pre-vendor';
```
sherakama: Adding a pre-vendor option is important as bourbon/neat/bitters can require some variables to be set BEFORE they are imported.

scss/elements/helpers/_layout.scss
```diff
@@ -0,0 +1,45 @@
+@charset 'UTF-8';
+
+///
+/// LAYOUT COMPONENT
+///
+
+// Flexbox positioning to move side navigation below main content on small screens.
```
sherakama: Shouldn't this be in the sidenav then?

kgcreative:
Dev standards are still using old bourbon/neat right? Completely different
syntaxes and lots of potential deprecation

sherakama:
Aye, they are, and yes, there is. A few items will have to be re-factored to upgrade so it can come later.

PR #100, opened 2018-04-19
Quote element

AlishaOber:
Created quote element within elements directory based on design Kerri made for the Engineering project.


josephgknox:
This looks really great! A few of things/thoughts:

1. Were identifying Elements as (for the most part) HTML items. That said, for how you've built and imagine using the quote, might it be a component rather than an element?
2. We won't want our Elements or Components generating markup, so let's make a quote mixin (see other elements and components for examples)
3. Let's use available mixins where we can for easier updating and overriding in partials. So, if we're overriding the media block image margins, let's do it with the @include margin() mixin.
4. Question: do we always want the quote only taking up half of the grid system? If not, should we remove the decanter-width-one-half from its wrapping <div>
5. Question: for semantics, should the quote be wrapped in an <h3>? Wondering if instead we assign it styles individually (wrapped in a <p> tag or separate <div> and assign it H3 styles. Open to thoughts here...

JBCSU:
I do think quote should be an element, but I'm wondering if we actually want 2 elements: <blockquote> and <q>? I think what we have so far fits better with <blockquote>, and should be marked up as such. Should it be renamed as blockquote to make it clear which element we're referring to?

I agree that the text should not be enclosed in an <h> tag. It should receive the base styling for the <blockquote> element.

templates/elements/quote/quote.html.json
```diff
@@ -0,0 +1,7 @@
+{
+  "modifier_class": "decanter-width-one-half",
```
sherakama: @josephgknox, The one half width is provided here in the example data an is not a requirement.

josephgknox: Right, agreed. The example (because it's one quote by itself) won't be impacted too much without it. For consistency, I'm wondering if not including grid wrappers when not necessary will ensure that our sample markup in the library doesn't include unnecessary stuff. I'm cool either way, though.

josephgknox:
@JBCSU thanks, JB! We faced the same question with our button. It was originally an element, but we moved it to the component (atom) section. There are going to be some base-level items such as these that fit in multiple boxes. Perhaps we have fallback styles for `<blockquote>`, too, then in this case. That's the route we went with buttons. There is a `decanter-button` class, which has styles applied to it and to the `<button>` element. I'm OK with keeping this in the Elements section if you both prefer.

AlishaOber:
@josephgknox I debated where this should live; my original thought was within the molecule component. I'm open to moving it elsewhere if there is a consensus. 

Based on the comments, it seems like I should do the following: 
1) For the example, just leave the modifier class empty. It doesn't seem like there's any reason for it not to be 100% width (except that the design wasn't). 
2) Change out the <h> tags and style <p> tags instead. In that case I'll likely no longer use the media-body element and style specifically for the quote element. I'll take a look at the mixins for this.

Did I miss anything?

josephgknox:
@AlishaOber all sounds good! A third item would be to create a mixin for the quote element/component and `@include` it on the`decanter-quote` class.

I also debated this when building out buttons, and ended up moving them from the Elements to the Components section. Again, I am game either way, but think we should be consistent with whatever we collectively decide. So, if we vote to keep the Quote in Elements because of `<blockquote>`, we might also want to do the same for buttons because of `<button>`.

Most important to wrapping this one up is the two items you've identified plus the third that I've noted here. Organizationally, we can move the Quote later if we so choose. You can choose where you think is best now.

Hope this helps!

josephgknox:
@AlishaOber sweeeet! Merging.

PR #104, opened 2018-04-23
Icon postcard block completed with approval from John on the design

AlishaOber:


josephgknox:
Minor tweak to margin call in mixin for consistency; removed two extra spaces in template file. This looks good. Merging.

josephgknox:
@sherakama let me know if you're good with merging, too.

templates/components/molecules/icon-block/icon-block.html.twig
```diff
@@ -0,0 +1,10 @@
+<div class="decanter-icon-block {{ classes }}">
+  <img class="decanter-icon-block__img" src="../../../img/circle.png" alt="Please provide alt text">
```
sherakama: Please update the src attribute to include a template variable so we can add our own src files when we use this template elsewhere. 

Please also do the same with the alt text.

templates/elements/quote/quote.html.twig
```diff
@@ -1,6 +1,6 @@
 <div class="decanter-quote">
-  <img class="decanter-quote__img " src="{{ image }}" alt="{{ altText }}">
-  <div class="decanter-quote__body ">
+  <img class="decanter-quote__img" src="{{ imageSource }}" alt="{{ alt }}">
```
sherakama: @josephgknox 

Do not use camelCase for variable names

The standard is to use underscore. See: https://twig.symfony.com/doc/2.x/coding_standards.html

change imageSource to `image_source` or better yet the shorter `img_source` or even shorter `src`

josephgknox: For sure! Thanks! Fixed.

josephgknox:
@sherakama LMK if you think this is ready to merge.

css/decanter.css
```diff
@@ -4,26 +4,34 @@ html {
 *, *::after, *::before {
   box-sizing: inherit; }
 
-/*! normalize.css v8.0.0 | MIT License | github.com/necolas/normalize.css */
+/*! normalize.css v6.0.0 | MIT License | github.com/necolas/normalize.css */
```
sherakama: Looks like normalize got rolled back to v6 here. I will update and re-compile.

templates/elements/quote/quote.html.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "Jane Doe",
-  "bio": "Jane Doe's title<br/>Jane Doe's company",
-  "text": "This is the actual quote. Lorem ipsur adipisicing elit, sed do eiusmod tempor incididunt ut labore et...",
-  "image": "../../../img/headshot.jpg",
-  "altText": "Jane Doe"
+  "name": "John Doe",
+  "detail1": "Detail 1",
```
sherakama: @josephgknox There are no detail variables in this template. What are they used for?
Also, bio is now missing.

josephgknox:
💯

PR #105, opened 2018-04-23
Quote tweaks

JBCSU:
I made a few changes to the `quote.html.twig` template to make it easier to use directly in a theme. Specifically, I'm using variables for the path to the image and to the image alt text, rather than hardcoding those.

I also renamed the `.decanter-quote__detail` class to `.decanter-quote__bio` to make it (hopefully!) clearer that the detail pertained to the person being quoted, not the quote itself.


templates/elements/quote/quote.html.twig
```diff
@@ -1,8 +1,8 @@
 <div class="decanter-quote">
-  <img class="decanter-quote__img " src="../../../img/circle.png" alt="Please provide alt text">
+  <img class="decanter-quote__img " src="../../../img/{{ image }}" alt="{{ altText }}">
```
sherakama: @JBCSU Let's change the src attribute to `src="{{ image }}"` so we can supply full paths to this template instead of assuming `../../../img/`

JBCSU:
Yes, of course. Great point!

templates/elements/quote/quote.html.twig
JBCSU: Done.

josephgknox:
Looks great!

PR #106, opened 2018-04-24
update neat-omega, grunt, grunt-sass, jquery, and associated dependencies

JBCSU:
I used [nvm](https://github.com/creationix/nvm) to set v9.10.1 of node (5.6.0 of npm) as my default version. It turned out I didn't have grunt installed for that version, so I (re-)installed grunt for 9.10.1, then did `npm update`. This updated the neat-omega, grunt, grunt-sass and jquery npm modules, and scores (hundreds?) of their dependencies. Everything seems to be building properly for me. Is it ok to update to more recent versions of these modules, or should I use an earlier version of node / npm? If you prefer that I back out my changes, what version of node / npm are y'all using? I think I was using 8.9.4 of node before (still 5.6.0 of npm), but I'm not positive.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


sherakama:
I am fine with these updates. I have found that different versions of node have caused different versions of the dependencies to appear in this build. We might consider putting a node version dependency in the package.json to ensure more uniformity.

josephgknox:
I'm also fine with these updates. And, related to our meeting earlier today, (so long as within reason) am 100% game for ensuring we have the latest and greatest of all the things!

I'll merge.

PR #108, opened 2018-04-26
Specify node version

### LINKED ISSUE #107: Create .nvmrc file for node consistency
Please create a `.nvmrc` config file with the version of node that should be used when working with decanter and nvm.

Please set it to the version you recently compiled with.

Thanks,

JBCSU:
Close #107 - Specify a node version number to be used with Decanter.

Per [nvm's suggestion](https://github.com/creationix/nvm#deeper-shell-integration), I also installed and activated [avn](https://github.com/wbyoung/avn), and verified that it does auto-switch my version of node when I cd to the directory. Nice!


josephgknox:
I'm only familiar with .nvmrc files to the extent of the documentation guide linked to on the related task. I've looked through it (in all its simplicity) and all here looks good; version is specified.

@sherakama verifying that you have OK'd this, too?

PR #109, opened 2018-04-27
README update with installation and usage instructions

AlishaOber:


README.md
```diff
@@ -15,3 +15,74 @@ Accessibility
...
 This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://ucomm.stanford.edu/policies/accessibility-policy.html](https://ucomm.stanford.edu/policies/accessibility-policy.html).
+
+System Requirements
+---
+
+To build and deploy code, you must have the following installed on your system:
+
+* Node
+* Grunt
+
```
sherakama: We might want to add KSS to this until we can get it from being a global install.

README.md
```diff
@@ -15,3 +15,74 @@ Accessibility
 
 Evaluation Date: 201X-XX-XX  
 This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://ucomm.stanford.edu/policies/accessibility-policy.html](https://ucomm.stanford.edu/policies/accessibility-policy.html).
+
+System Requirements
+---
+
+To build and deploy code, you must have the following installed on your system:
+
+* Node
```
sherakama: Perhaps also add version numbers?

README.md
```diff
@@ -15,3 +15,74 @@ Accessibility
...
+* If you run into any issues with some of the tasks run, npm rebuild and try again.
+`npm rebuild`
+* Also make sure that git is installed globally as some bower packages require it to be fetched and installed.
+
+Development Workflow
+---
+
+### One-time Setup
+
+The following steps need to be performed once to get your local environment up and running.
```
sherakama: Lando is optional and not a requirement. I do not want to force anyone to use this webserver if they don't want to. Perhaps we could change this section to an optional install and move the KSS setup to the Installation section?

README.md
```diff
@@ -15,3 +15,74 @@ Accessibility
...
+* `cd` to the local directory where you keep your git repos
+* `git clone https://github.com/SU-SWS/decanter.git`
+* `cd decanter`
+* `npm install -g kss`
+* `npm install` to install your remaining dependencies
+
+Workflow
+---
+
+### Starting Your Environment
```
sherakama: We can keep the lando stuff in this section but we should move the styleguide and css compilation to its own section.

README.md
```diff
@@ -15,3 +15,74 @@ Accessibility
...
+
+* Start Docker by double clicking the app icon.
+* Start Lando by double clicking the app icon.
+* Initialize Lando by typing `lando start`. For full Lando documention, see [https://docs.devwithlando.io/](https://docs.devwithlando.io/).
+* To build the styleguide, use `grunt styleguide`.
+* To compile the files, start the watch task, and automatically refresh the browser when changes are made, use `grunt`.
+
+### Development
```
sherakama: We have a wiki page for this. I like having the short notes here but we should add a link to the wiki page.

README.md
josephgknox: +1. The more documentation (and the more specific), the better here, I think. Let's be diligent about making sure we keep this updated as things change.

README.md
josephgknox: +1

README.md
josephgknox: I do think that having some basic, outline-like steps here is still useful and would allow for more advanced users to see what they need and proceed without having to navigate to the Wiki. That said, cleaning up the steps to make them even shorter would be a good idea (e.g., "Start Docker by double clicking the app icon." to "Start Docker.").

sherakama:
@josephgknox perhaps when you are working with JBC on integrating this package in to new work you can help finish off this documentation and PR.

PR #119, opened 2018-05-16
Lando db gone

sherakama:
@JBCSU 

Thanks for that pull request. That does fix the MySQL issue on OSX, but we are not using mysql for this project. Perhaps we just get rid of it so that it doesn't cause any more issues.

Here is a PR to remove the DB and just run Apache/PHP

Thoughts?

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,13 +1,20 @@
 name: decanter
-recipe: lamp
-config:
-  webroot: styleguide
-  php: '7.1'
-services:
+proxy:
   appserver:
+    - decanter.sws
+    - decanter.loc
+    - decanter.lndo.site
+# Spin up services to run the styleguide
+services:
+  # Create a php 7.1 instance served by the latest apache with https turned on
+  appserver:
+    type: php:7.1
+    via: apache
     ssl: true
-  database:
-    type: mysql:5.7
+    webroot: styleguide
+# Add some nice command routing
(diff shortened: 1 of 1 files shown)


JBCSU:
This is so much better than my lame workaround. Thanks!

Apologies for not taking the time to learn lando enough to figure out how to build without a db.

JBCSU:
Why do we need 3 appservers / url's?

sherakama:
Hey @JBCSU,

There is nothing lame about adding the MSYQL fix. It was a fix. We're talking code and there are many different solutions to our problems. This is just another one and a good time to talk about it and both of us learn. 

I figured getting rid of something we weren't using might be better long term. The trade off was that I had to add the proxy lines in which it gave us three urls. For some reason when I removed the LAMP recipe I was not getting my `decanter.sws` proxy to show up and only localhost was showing. I like my clean urls without port numbers so I put two of the conventions that we use for local development urls. 

I'd rather it have the original behaviour where it picks up my proxy settings from `~/.lando/` but I wasn't able to quickly figure out how to do that. Perhaps one of us can find the time to find out before this gets merged.

sherakama:
As to the composer error, it looks like it was a network issue. The setup processes were unable to load the remote data point and failed to complete. I have seen this often on the Stanford network and it is annoying but usually resolvable by simply trying again.

PR #170, opened 2018-07-09
Decanter 159 upgrade neat

bixgomez:
# READY FOR REVIEW

# Summary
- Upgrading Decanter to v3.0

# Steps to Test
- No real steps, aside from installing the new version and confirming that it throws no errors.

# Notes
- Currently, simple media queries are using Breakpoint-Sass.  We have decided to NOT use this resource, so this will change in favor of custom mixins to achieve the same goal.  However, I believe it would be most efficient to do this work in the context of my first layout ticket.
- Also, you will notice some new templates for the basic page layouts.  This lays the groundwork for my four upcoming tickets. 

- Pretty much everything!

- [JIRA ticket](https://stanfordits.atlassian.net/browse/UCP-471)


Gruntfile.js
```diff
@@ -5,9 +5,10 @@ module.exports = function(grunt) {
       options: {
         includePaths: [
           "node_modules/bourbon/core",
-          "node_modules/bourbon-neat/app/assets/stylesheets",
-          "node_modules/normalize.css",
+          "node_modules/bourbon-neat/core",
+          "node_modules/breakpoint-sass/stylesheets",
```
sherakama: I was hoping that we could avoid the addition of this library. Is this a stop gap with another round of refactoring coming or something that is absolutely needed?

sherakama:
@bixgomez Is all of this work in #171? If so, let's close this PR and work through the comments on the other issue.

sherakama:
Closing and pivoting to use grid/flexbox.

PR #173, opened 2018-07-13
added semibold and other source sans pro font styles that were missing

josephgknox:
# READY FOR REVIEW

# Summary
- Added missing Source Sans Pro font styles
- Address gaps between mocks and code output in projects

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Ensure that the new font styles (font-style and font-weight) are available (e.g., font-weight: 900)

- Decanter
- Archaeology

- N/A

# See Also


josephgknox:
Cc: @sherakama @JBCSU

josephgknox:
@JBCSU yep, there is a file (or two or three or four) for each of the font variations, and all font variations are downloaded (@font-face) even if they are not used (so that anytime a related variable or mixin is used it works).

I like your idea of showing all the fonts and their variations. I can work on getting that in.

sherakama:
I am going to add a backlog ticket to be able to dynamically load fonts so that not everything is loaded all the time. As fonts are one of the heaviest items this would be a nice improvement. Not for the score of this work however.

josephgknox:
@JBCSU added all style variations to the example template at Core > Fonts. Heading styles, paragraph styles, and all variations of the font can now be viewed. This space (as well as many others) can use some theming of the style guide itself to reduce spacing and margins between items.

@sherakama thanks for adding that! We're using three Google fonts, so we could use the @import rather than @font-face (so that we don't have to host the fonts at all) if we want to. In either case, the most bulletproof and least heavy option would be best.

PR #176, opened 2018-07-17
Small Tweaks and Fixes

josephgknox:
# READY FOR REVIEW

# Summary
- Very small changes, but are being added so that they don't need to be overrode in project partials every time
- removed margin-top from heading 1's
- removed the hard-coded font family on buttons

# Needed By (Date)
- As soon as possible (would unblock some project work)

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Ensure that tweaks are applied

- decanter
- stanford_basic
- suhumsci (theme)

- N/A

# See Also

### DIFF
--- a/css/decanter-grid.css
+++ b/css/decanter-grid.css
@@ -11,8 +11,7 @@ body {
   overflow-x: hidden; }
 
 .lt-ie9 * {
-  -webkit-filter: none !important;
-          filter: none !important; }
+  filter: none !important; }
 
 [hidden] {
   display: none !important; }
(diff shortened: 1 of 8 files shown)


josephgknox:
Cc: @bixgomez

sherakama:
@josephgknox Can you clean up the codeclimate comments before merging.

josephgknox:
@sherakama I'm feeling like we want to ignore these 4 CC issues, in this case. We would need to add a few extra (and arguably unnecessary) lines to the SCSS and nest deeper to address them. Thoughts?

sherakama:
@josephgknox all good. But do you really need `a.selector`? could you get away without the tag on the selector or does some orther css scope blow out the styles?

josephgknox:
@sherakama based on our conversation this morning (on Decanter vs Drupal/WP vs theme), I have made the related update (see last commit). This also addressed the CC issues.

PR #187, opened 2018-08-14
Decanter 161  grid version  v04

bixgomez:
# READY FOR REVIEW

# Summary
- Implementing functional Flexbox class-based grid system

# Steps to Test
1. Visit /section-core.html#kssref-core-grid in the style guide
1. Confirm that the grid performs as expected across all breakpoints, and in IE11

- [GitHub ticket](https://github.com/SU-SWS/decanter/projects/11#card-11056610)


josephgknox:
This looks great, and is very close. A couple of things (the first of which is bigger than the other):

- we talked about changing the class names in our meeting last week. Something like `decanter-span[#]`
- due to the new class-based system, which is being leveraged in some of the style guide's examples, the "Colors" section is borked. Looks like it just needs the `decanter-grid` wrapper

PR #209, opened 2018-09-02
151 footer rg

bixgomez:
# NOT READY

# Summary
- Implements a horizontal list mixin.

@sherakama : I created a horizontal list mixin, with a few preliminary arguments, and I have applied it to the footer menu.  I've tried to be as explicit as I could in my comments, so hopefully they make sense.  I got a little lost in the nomenclature -- couldn't decide whether to name the mixing horizontal-list, or list-horizontal.  I settled on the latter because it matches the existing list-unstyled mixin, which it incorporates.  Since this has been my last week on the project, I'll leave it to you to decide how to move forward, but I am interested in your feedback.


sherakama:
Closing and switching to branch `151-Footer`

PR #212, opened 2018-10-04
133 CTA Molecules

yvonnetangsu:
This is WIP - only worked on 2 out of the 3 variants. Plan on cleaning up code for the 2 existing ones and perhaps adding options for tinting/shading the image. 

The class name uses singular "cta" eg, ".decanter-cta" but I did use plural "ctas" for some files, eg, ctas.twig, ctas.json - I was trying to be consistent with naming for eg, alerts.twig etc. I'd be happy to change all those to singular "cta".


sherakama:
Thanks for this Yvonne,

On initial review, there are a couple of changes that I think could improve the maintainability of this. Currently, there are three variants of the CTA in three mixins with a fair amount of common code. I would love this if it could be refactored so there was one mixin and 4 css classes for the variants. One common CTA mixin that provides the base styles for what a mixin looks like and then 3 css classes which add on top of the common CTA code to produce their version.

yvonnetangsu:
That makes total sense, Shea! Thanks for the suggestion - will do.

yvonnetangsu:
Would it make more sense for me to wait to revise this after we have the Decanter meeting next Mon/Wed? I heard from John yesterday that the class naming is going to be different.

sherakama:
Feel free to revise now. No need to wait.

yvonnetangsu:
Just refactored the code and added the variant with the icon. I wasn't sure what's the best way to add something in twig that's only in some modifier classes, in this case, the icon - so I add a "display:none" to the default mixin (for variants with no icons). I'm sure you have a better way of doing it - let me know. Thanks!

core/scss/components/molecules/_ctas.scss
```diff
@@ -0,0 +1,63 @@
+@charset "UTF-8";
```
sherakama: I like your point on how to name these files and that you followed the alerts pattern. However, I think that we should change and make all of these patterns singular and remove the plural names. So this file should be `cta`.

core/scss/components/molecules/_ctas.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.decanter-cta--button-bottom,
+.decanter-cta {
+  @include cta--button;
+}
+
+.decanter-cta--button-bottom-icon {
+  @include cta--button;
+
+  @import "https://use.fontawesome.com/releases/v5.4.1/css/all.css";
```
sherakama: We have some discovery next week about how to include this in the Decanter project so in the mean time, please remove this.

core/scss/components/molecules/_ctas.scss
```diff
@@ -0,0 +1,63 @@
...
+// Style guide: Molecules.CTAs
+//
+
+.decanter-cta--button-bottom,
+.decanter-cta {
+  @include cta--button;
+}
+
+.decanter-cta--button-bottom-icon {
+  @include cta--button;
```
sherakama: As this is an additive css class, you won't need this mixin here.

core/scss/components/molecules/_ctas.scss
```diff
@@ -0,0 +1,63 @@
...
+    transform: translate(-50%, $_shift);
+    color: #fff;
+    font-size: 4em;
+    text-shadow: 0 0 10px rgba(0, 0, 0, .4);
+  }
+
+}
+
+.decanter-cta--button-center {
+  @include cta--button;
```
sherakama: You shouldn't need this here.

core/templates/components/molecules/ctas/ctas.json
```diff
@@ -0,0 +1,8 @@
+{
+  "grid": "grid--4-of-12",
```
sherakama: Do you expect others to be able to change this?

core/templates/components/molecules/ctas/ctas.twig
```diff
@@ -0,0 +1,5 @@
+<a {{ attributes }} href="{{ link|default("https://stanford.edu") }}" class="{{ attributes.addClass('decanter-cta') }} {{ modifier_class }} {{ grid }}">
```
sherakama: You shouldn't have to add the class to attributes like this. Instead, just have the class text. eg:
`<a{{ attributes }} href="{{ link|default("https://stanford.edu") }}" class="decanter-cta {{ modifier_class }} {{ grid }}">`

sherakama: no space after the `<a` and `{{ attributes }}`

core/scss/components/molecules/_ctas.scss
yvonnetangsu: Changed all file names and references to singular cta

core/scss/components/molecules/_ctas.scss
yvonnetangsu: Sounds great - removed.

core/scss/components/molecules/_ctas.scss
yvonnetangsu: Removed.

core/scss/components/molecules/_ctas.scss
yvonnetangsu: Removed.

core/templates/components/molecules/ctas/ctas.json
yvonnetangsu: This is the part I'm unsure about how to handle - elements that are not 100% width/things that would probably occupy a certain number of grid columns depending on how the site builder wants it. Let me know how I should structure this so it make sense. Thanks!

core/templates/components/molecules/ctas/ctas.twig
yvonnetangsu: Changed - thanks!

Issue #184, opened 2018-08-08
Deprecated import of normalize.css


JBCSU:
**Describe the bug**
`decanter.scss` imports `normalize` from the [normalize.css](https://www.npmjs.com/package/normalize.css) npm package. That's a `.css` file, not `.scss`. webpack's sass loader warns that "Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass."

**To Reproduce**
1. Configure webpack to use `sass-loader` to process a .scss file which includes `@import "decanter";`. And good luck to you doing that.
2. Use webpack to generate your css from your scss.
3. See this warning:

```DEPRECATION WARNING on line 21, column 8 of ./node_modules/decanter/scss/decanter.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.
```

**Expected behavior**
`decanter.scss` should pre-process without warnings.

**Proposed solution**
Find and use an appropriate `normalize.scss`, perhaps [JohnAlbin's](https://www.npmjs.com/package/normalize-scss)?

sherakama:
We will remove the npm dependency on normalize.css and hardcode into our project.

JBCSU:
I tried replacing the `normalize.css` package with [normalize-scss](https://github.com/JohnAlbin/normalize-scss), but that resulted in a conflict on `$base-font-size`: `normalize-scss` defines that variable in px, but we use it in calculation(s) that refer to variables defined in rem, resulting in a unit conflict error when compiling our scss. It was not straightforward to resolve this error. Also, normalize-scss is a version behind normalize.css (7.0.0 vs. 8.0.0). So we decided to remove `normalize.css` from our npm dependencies, rename `normalize.css` as `_normalize.scss`, save it in our scss source directory, and import it from there. We understand and agree that this means we'll have to manually update normalize as new versions become available.

PR #239, opened 2018-10-30
resolve 184: stop importing .css file into .scss

JBCSU:
# READY FOR REVIEW

# Summary
- TL;DR stop importing `normalize.css` into our `.scss`, as that functionality is deprecated

# Needed By (Date)
- Before webpack stops supporting importing `.css` files

# Urgency
- Low

# Steps to Test
1. `npm install` to remove the dependency on `normalize.css`
1. `grunt styleguide` to rebuild our .css and the styleguide
1. Verify that nothing has changed. It may be easiest to confirm that the first 143 lines of the new `decanter.css` are the same as the first 299 lines of the old one (removed all block style comments for CodeClimate).

- N/A

- #184 

# See Also

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -76,3 +76,4 @@ exclude_patterns:
 - "styleguide"
 - "kss/**/*"
 - "core/js"
+- "core/scss/core/_normalize.scss"
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,7 +6,6 @@ module.exports = function(grunt) {
         includePaths: [
           "node_modules/bourbon/core",
           "node_modules/bourbon-neat/app/assets/stylesheets",
-          "node_modules/normalize.css",
           "node_modules/neat-omega/core",
           "node_modules",
           "core/scss"
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -1,38 +1,16 @@
-/*! normalize.css v8.0.0 | MIT License | github.com/necolas/normalize.css */
-/* Document
-   ========================================================================== */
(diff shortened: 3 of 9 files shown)


JBCSU:
@sherakama I have 3 CodeClimate issues remaining for the new `_normalize.scss` file:

1. Line 58: "border-bottom: 0 is preferred over border-bottom: none" - super easy to fix
1. Line 60: "Property text-decoration already defined on line 59" - easy to fix, but I wonder why the property was defined twice in the first place?
1. Line 57: "Avoid qualifying attribute selectors with an element" - I don't think there's any way around this without changing the semantics.

It turns out all of the issues are in this block:

```css
abbr[title] {
  border-bottom: none; // 1
  text-decoration: underline; // 2
  text-decoration: underline dotted; // 2
}
```

I don't think we're ever going to emit an `<abbr>` tag, and we're even less likely to emit one with a `title` attribute. What do you think about just deleting that block from normalize? If you don't want to delete it, is there a way to retain the semantics without using an attribute qualification? If not, I can fix the other 2 errors, but I think we'll get that CodeClimate error on line 57 for ever PR for the rest of time.

sherakama:
As this file is technically a library and we don't lint external libraries, I think we can be safe to ignore this file from codeclimate scans. Feel free to put an ignore path in to the `.codeclimate.yml` file for this.

JBCSU:
@sherakama I've reverted `_normalize.scss` to vanilla version and CodeClimate is successfully ignoring it. Ready to merge?

sherakama:
Wonderbar.

Thanks.

Issue #223, opened 2018-10-24
Create brand bar component.


sherakama:
Create the top brand bar component as found at the top of most Stanford websites.

PR #243, opened 2018-10-30
223 brand bar

yvonnetangsu:
# READY FOR REVIEW
- A first pass but would like a review as a sanity check

# Summary
- Global brand bar component with 4 variants

# Needed By (Date)
- Sometime this week before Friday if possible

# Urgency
- Not urgent

# Steps to Test

1. Check out this branch.
1. Run grunt styleguide
2. Click on "Components" and look at section 3.2 "Brand Bar" and check that the 4 variants match the design in the Figma document Stanford Design Library (under "Header & Footer" -> "Brand Bar")

- Future Drupal and Wordpress templates

- JIRA ticket https://github.com/SU-SWS/decanter/issues/223
- It makes use of the CSS grid mixin and is dependent of the current breakpoints


core/scss/components/brand-bar/_brand-bar--dark.scss
```diff
@@ -0,0 +1,5 @@
+@charset "UTF-8";
+
+#su-brand-bar.su-brand-bar--dark {
```
sherakama: See comments for bright.

core/scss/components/brand-bar/_brand-bar--white.scss
```diff
@@ -0,0 +1,11 @@
+@charset "UTF-8";
+
+#su-brand-bar.su-brand-bar--white {
```
sherakama: Please leave a comment and remove the ID.

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,32 @@
...
+// .su-brand-bar--bright - White wordmark on bright red background
+// .su-brand-bar--dark - White wordmark on dark grey background
+// .su-brand-bar--white - cardinal red wordmark on white background
+//
+// Markup: ../../templates/components/brand-bar/brand-bar.twig
+//
+// Style guide: Components.Brand Bar
+//
+
+#su-brand-bar,
```
sherakama: Please remove the ID.

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,32 @@
...
+.su-brand-bar--default {
```
sherakama: The default version does not need a modifier class. It can be simply `su-brand-bar`

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  @include centered-grid-container;
+  width: 100%;
+  height: 30px;
+  background-color: $color-cardinal-red;
+
+  .su-brand-bar__logo {
```
sherakama: This can be popped out of the parent wrapper and be on it's own if it is the default item. This makes it easier to override in variants.

core/templates/components/brand-bar/brand-bar.twig
```diff
@@ -0,0 +1,3 @@
+<section{{ attributes }} id="su-brand-bar" class="{{ modifier_class }}">
+  <div class="su-brand-bar__container"><a{{ attributes }} class="su-brand-bar__logo" href="https://stanford.edu">Stanford University</a></div>
```
sherakama: I think you can remove the attributes variable from the `<a>` tag or at the very least, change the variable name so that the `<section>` wrapper attributes do not get applied to the `<a>` tag.

package-lock.json
```diff
@@ -16,10 +16,10 @@
       "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
```
sherakama: Is there anything in this lock file that is needed for this PR? If not, I would rather see this not committed. To revert you can easily recover the version in the master branch with `git checkout master -- package-lock.json`

core/scss/components/brand-bar/_brand-bar--bright.scss
```diff
@@ -0,0 +1,5 @@
+@charset "UTF-8";
+
+#su-brand-bar.su-brand-bar--bright {
```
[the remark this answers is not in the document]
josephgknox: I, too, would like to see the ID removed, if possible.

core/scss/components/brand-bar/_brand-bar--white.scss
```diff
@@ -0,0 +1,11 @@
...
+  background-color: $color-white;
+
+  .su-brand-bar__logo {
+    color: $color-cardinal-red;
+    -webkit-font-smoothing: auto;
+    -moz-osx-font-smoothing: auto;
```
josephgknox: We have a font-smoothing mixin. Would that one work for you? If not, perhaps the mixin can be refactored with variables so that we can pass through whatever value we want. Right now it is:

`@mixin font-smoothing {
  -moz-osx-font-smoothing: grayscale;
  -webkit-font-smoothing: antialiased;
}`

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,32 @@
...
+// Markup: ../../templates/components/brand-bar/brand-bar.twig
+//
+// Style guide: Components.Brand Bar
+//
+
+#su-brand-bar,
+.su-brand-bar--default {
+
+  @include centered-grid-container;
+  width: 100%;
```
josephgknox: We don't want our components generating markup if they don't have to. We also want to make our mixins available to those using the `decanter-no-markup` file. So, I'm thinking we should package the code up into a mixin and call the mixin at the class level.

core/scss/components/brand-bar/_brand-bar--bright.scss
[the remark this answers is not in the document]
yvonnetangsu: Yup, only reason I did that was because I defined the default styles inside the #su-brand-bar {}, so the variants need that id selector in front to override those settings. I think I'll restructure it so that the default styles are defined in the default class, and not inside the ID {}.

And I'll add comments on the variant files.

core/scss/components/brand-bar/_brand-bar--dark.scss
yvonnetangsu: see above reply

core/scss/components/brand-bar/_brand-bar--white.scss
yvonnetangsu: see above reply

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: see above reply

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: I think initially I did that and the html output by kss node was a little strange so I added the default class. Let me tweak it and try again and see what it gives me.

core/templates/components/brand-bar/brand-bar.twig
yvonnetangsu: Sure, will do. I mistook the {{attributes}} tag as a generic "the site builder can add any attribute here" tag.

package-lock.json
yvonnetangsu: Sure, will get it out.

core/scss/components/brand-bar/_brand-bar--white.scss
yvonnetangsu: I was looking into antialiasing when working on the homesite. The default setting works well on light text over darker background (so for the 3 other variants), but for cardinal red wordmark over white, it kind of needs the opposite setting (which is what I'm adding in this file) for the font weight to look the same in all variants. Otherwise the wordmark looks thinner for this variant. Let me know if there's a better way of handling this.

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: I could rewrite it so it's similar to how I wrote the CTA mixin.

package-lock.json
yvonnetangsu: And thanks for the git command - that's useful to know!

core/scss/components/brand-bar/_brand-bar--white.scss
yvonnetangsu: And by "default setting", I meant the font smoothing styles that's currently being bundled into the wordmark logo component which has the same settings as in the font-smoothing mixin.

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: Sure! It's been taken out of parent wrapper in default scss file.

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: Disregarding this comment after discussion with Joe and Shea.

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,36 @@
...
+//
+// Markup: ../../templates/components/brand-bar/brand-bar.twig
+//
+// Style guide: Components.Brand Bar
+//
+
+.su-brand-bar {
+  @include centered-grid-container;
+  width: 100%;
+  height: 30px;
```
JBCSU: Should the height of the brand bar be hardcoded in px? The wordmark is text, rendered in the ligature font. Should the height of the container adjust if the user changes font size? Or is our brand bar fixed height? If the latter, should the font-size be specified? Right now it's inheriting from `<body>`.

core/templates/components/brand-bar/brand-bar.twig
```diff
@@ -0,0 +1,19 @@
...
+ *   - Bright: White wordmark over bright red background
+ *   - Dark: White wordmark over dark grey background
+ *   - White: Cardinal red wordmark over white background
+ *
+ * Available variables:
+ * - attributes: For additional HTML attributes not already provided.
+ * - modifier_class: Additional css classes to change look and behaviour.
+ */
+#}
+<section id="su-brand-bar" class="su-brand-bar {{ modifier_class }}">
```
JBCSU: Homesite currently has a `<section id="brand-bar">`, but that section contains several items:

- The wordmark
- The gateway nav
- The menu toggle (mobile)
- The search toggle (mobile)

Looking at the candidate patterns for the masthead, it looks like the brand bar will only ever contain the wordmark, even on mobile. Does it therefore need to be wrapped in a `<section>`. Could we just use the `<div>`? I anticipate there will be a `<section>` for the masthead. Maybe we don't need to nest an additional section that only contains the word mark?

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: I hardcoded it in px since it seems like it will always be 30px tall on any devices - thought that might be less room for surprises. I should definitely specify a font size....I guess that isn't dependent on anything (breakpoint eg) so I'll just hardcode it in. Thanks!

core/templates/components/brand-bar/brand-bar.twig
yvonnetangsu: I think the <section> tags are needed (or at least an outside container tag pair) to work with Richard's CSS grid mixin. The mixin takes the first element inside that parent outside container and stick it in the 2 column of the grid. In the beginning, I didn't have the <div> container around the <a> element, and it ended up making the whole brand bar clickable (the whole center container portion) because it sticks the <a> element inside the 2nd grid column and it takes up the full width of that column.

JBCSU: Ah. I see. Thanks!

You may end up refactoring this when you work on the masthead, then. I suspect the masthead will end up in a `<section id="masthead">`, and maybe that will `@include centered-grid-container`. We'll see what makes sense then.

yvonnetangsu: Yeah, I can definitely see some refactoring needed.

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: Just added font size and pushed it up.

JBCSU: Should the `font-size` go in the `logo` mixin with all the other font specifications?

JBCSU: I see the `logo` mixin is also in the global footer, so you probably did put it in the right place. Sorry for my confusion

yvonnetangsu: I think I prefer it outside of the logo mixin since the font size of the ligature font is going to change depending on where it is used (eg, in lockup or footer it's going to be different).

core/scss/components/brand-bar/_brand-bar.scss
sherakama: Components are allowed to generate markup.

core/templates/components/brand-bar/brand-bar.twig
sherakama: only one attributes variable can be sent to this twig file so it should not be used on more than one html element unless you intend to have exactly the same properties.

yvonnetangsu: Yup, I've removed it from `<section>` so only `<a>` has the {{ attributes }} variable now. Thanks.

core/scss/components/brand-bar/_brand-bar--bright.scss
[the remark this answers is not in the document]
josephgknox: Looks good now!

core/scss/components/brand-bar/_brand-bar--white.scss
josephgknox: I pushed up a commit with an update to the font-smoothing mixin (I updated the related call in this file, which outputs what we want still). LMK your thoughts.

core/scss/components/brand-bar/_brand-bar.scss
josephgknox: Yep! Thanks, all, for the discussion on this earlier in the week. Makes sense.

josephgknox:
This looks good to me.

core/scss/components/brand-bar/_brand-bar--white.scss
```diff
@@ -0,0 +1,12 @@
...
+// Brand bar with cardinal red wordmark on white background
+
+.su-brand-bar--white {
+  background-color: $color-white;
+
+  .su-brand-bar__logo {
+    @include font-smoothing(auto);
+    color: $color-cardinal-red;
+  }
+}
```
JBCSU: The white variant looks **really bad** on my large, non-retina monitor at home. See attached image. Is there anything you can do to clean up Stanford's wordmark so it looks excellent even on lower res monitors?

brand bar non-retina

core/scss/components/brand-bar/_brand-bar--white.scss
sherakama: This looks good.

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -0,0 +1,37 @@
+@charset "UTF-8";
+
+//
+// Brand Bar
+//
+// Stanford brand bar with the wordmark logo that links to the Stanford homepage. The brand bar sits at the very top of a website.
+//
+// .su-brand-bar - Default brand bar with white wordmark on cardinal red background
```
sherakama: You do not need to call out the default variant as the templates will handle rendering it.

sherakama: Remove this line.

core/scss/components/brand-bar/_brand-bar--white.scss
sherakama: eek

core/scss/components/brand-bar/_brand-bar.scss
yvonnetangsu: Done.

Issue #249, opened 2018-10-31
CTA Component Icon Variant


sherakama:
Change FA icon to a variable in the twig template where the user can add an icon image or svg.

Issue #133, opened 2018-05-23
CTA component (twig and sass)


sherakama:

josephgknox:
<img width="824" alt="molecule_image-cta" src="https://user-images.githubusercontent.com/8933670/44177610-8ff34b80-a0a3-11e8-882a-4544c9a9feed.png">

PR #251, opened 2018-11-01
249: Change CTA icon to a variable

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Changed the icon to a variable (icon_src) since FontAwesome integration isn't ready; also pulled some CSS styles of components out of parent class wrapper in the mixin file.

# Needed By (Date)
- End of week

# Steps to Test

1. Pull this branch
1. Run grunt styleguide and check that the 3 CTA variants are there and the icon one is rendering a svg icon from the img directory.
2. Check that the cta.twig file is set up correctly

- Decanter

- https://github.com/SU-SWS/decanter/issues/133
- https://github.com/SU-SWS/decanter/issues/249

### DIFF
--- a/core/scss/components/cta/_cta.scss
+++ b/core/scss/components/cta/_cta.scss
@@ -25,17 +25,17 @@
 
   .su-cta__icon {
     display: block;
+    max-width: 30%;
     position: absolute;
     top: 50%;
     left: 50%;
     $_btn-height: $lead-font-size + 3rem; //height of the button
     $_shift: calc(#{-50%} - #{$_btn-height / 2}); //calculate the extra shift up for the icon to compensate for the button height
     transform: translate(-50%, $_shift);
-    color: #fff;
-    font-size: 4em;
-    text-shadow: 0 0 10px rgba(0, 0, 0, .4);
+    //color: #fff;
+    //font-size: 4em;
+    //text-shadow: 0 0 10px rgba(0, 0, 0, .4);
   }
-
 }
 
 .su-cta--button-center {
--- a/core/scss/components/cta/_cta.scss
+++ b/core/scss/components/cta/_cta.scss
@@ -32,9 +32,6 @@
     $_btn-height: $lead-font-size + 3rem; //height of the button
     $_shift: calc(#{-50%} - #{$_btn-height / 2}); //calculate the extra shift up for the icon to compensate for the button height
     transform: translate(-50%, $_shift);
-    //color: #fff;
-    //font-size: 4em;
-    //text-shadow: 0 0 10px rgba(0, 0, 0, .4);
   }
 }
 
--- a/core/scss/utilities/mixins/cta/_cta--button.scss
+++ b/core/scss/utilities/mixins/cta/_cta--button.scss
@@ -26,10 +26,8 @@
   }
 
   .su-cta__img {
-    @include image-defaults;
     width: 100%;
     z-index: 0;
-    will-change: transform;
     transition: all .3s ease-out;
   }
 
--- a/core/scss/utilities/mixins/cta/_cta--button.scss
+++ b/core/scss/utilities/mixins/cta/_cta--button.scss
@@ -7,19 +7,12 @@
 //
 // Style guide: Mixins.CTA.cta--button
 //
+
 @mixin cta--button {
   display: flex;
   flex-direction: column;
   overflow: hidden;
   line-height: 0;
-
-  .su-cta__img {
-    @include image-defaults;
-    width: 100%;
-    z-index: 0;
-    will-change: transform;
-    transition: all .3s ease-out;
-  }
 
   &:hover {
     .su-cta__img {
@@ -32,6 +25,14 @@
     }
   }
 
+  .su-cta__img {
+    @include image-defaults;
+    width: 100%;
+    z-index: 0;
+    will-change: transform;
+    transition: all .3s ease-out;
+  }
+
   .su-cta__button {
     @include button-big;
 
@@ -41,7 +42,6 @@
       width: 100%;
       z-index: 100;
     }
-
   }
 
   .su-cta__icon {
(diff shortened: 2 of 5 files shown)


core/scss/components/cta/_cta.scss
```diff
@@ -25,17 +25,17 @@
...
     position: absolute;
     top: 50%;
     left: 50%;
     $_btn-height: $lead-font-size + 3rem; //height of the button
     $_shift: calc(#{-50%} - #{$_btn-height / 2}); //calculate the extra shift up for the icon to compensate for the button height
     transform: translate(-50%, $_shift);
-    color: #fff;
-    font-size: 4em;
-    text-shadow: 0 0 10px rgba(0, 0, 0, .4);
+    //color: #fff;
```
sherakama: If you don't need something please remove it. Don't leave unneeded code behind.

core/templates/components/cta/cta.twig
```diff
@@ -1,5 +1,31 @@
+{#
+/**
+ * @file
+ * CTA Component.
+ *
+ * Call to action component with clickable image and button.
+ * Variants:
+ *   - Default - Button at the bottom: With full-width button under the image
```
sherakama: I'm open for discussion on this. Do we really need to call out the default variant?

core/templates/components/cta/cta.twig
```diff
@@ -1,5 +1,31 @@
...
+ *   - Button in the center: With button in the center of the clickable image
```
sherakama: It would be nice if the CSS class or mixin was provided in the comment.

core/templates/components/cta/cta.twig
```diff
@@ -1,5 +1,31 @@
...
+ *   - Button at the bottom with icon in the center: With full-width button under the image and an icon in the center of the image
+ *
+ * Available variables:
+ * - attributes: For additional HTML attributes not already provided.
+ * - modifier_class: Additional css classes to change look and behaviour.
+ * - src: The href property of the image.
```
sherakama: This is now icon_src. Thank you.

core/templates/components/cta/cta.twig
```diff
@@ -1,5 +1,31 @@
...
+ */
+#}
 <a{{ attributes }} href="{{ link|default("https://stanford.edu") }}" class="su-cta {{ modifier_class }}">
-  <img class="su-cta__img" src="{{ src|default("../../../img/hero-banner.png") }}" alt="{{ alt|default("Please provide alt text") }}">
-  <i class="su-cta__icon {{ icon|default("fas fa-cat") }}"></i>
+  <img class="su-cta__img" src="{{ src|default("../../../img/hero-banner.png") }}" alt="{{ alt|default("Please provide alt text") }}" />
+
+  {# CTA icon #}
+  {% if icon_src is not empty %}
+    <img src="{{ icon_src }}" class="su-cta__icon" />
```
sherakama: This would benefit from an alt tag and `role="presentation"` for accessibility.

core/scss/utilities/mixins/cta/_cta--button.scss
```diff
@@ -32,6 +25,14 @@
     }
   }
 
+  .su-cta__img {
+    @include image-defaults;
+    width: 100%;
+    z-index: 0;
+    will-change: transform;
```
sherakama: This property is new to me. Thanks for the learning opportunity. At my first reading of the spec on https://developer.mozilla.org/en-US/docs/Web/CSS/will-change it indicates that this option should be used very sparingly. Could you add an SCSS comment to what you are trying to fix with this so I can better understand its need?

core/scss/utilities/mixins/cta/_cta--button.scss
```diff
@@ -32,6 +25,14 @@
     }
   }
 
+  .su-cta__img {
+    @include image-defaults;
```
sherakama: All image tags get this by default. Do you need this again for a particular style?

core/scss/components/cta/_cta.scss
yvonnetangsu: Sure - removed.

core/templates/components/cta/cta.twig
yvonnetangsu: I removed the "Default" part. I'm not sure if we need a default or if the button-bottom one should be the default, though I did code it that way (the default class without modifier class is the one with the button at the bottom).

core/templates/components/cta/cta.twig
yvonnetangsu: Added the CSS class names for the variants

core/templates/components/cta/cta.twig
yvonnetangsu: I renamed the variables so they are a bit clearer.

core/templates/components/cta/cta.twig
yvonnetangsu: Very true. Added.

core/scss/utilities/mixins/cta/_cta--button.scss
yvonnetangsu: Actually you're right - thank you for making me think about this more carefully. I learned about this property while looking at the Stanford homepage code and thought maybe it would make the animation a bit smoother, but I don't think it's doing much anyway and probably should be used only when really needed. It's now removed.

core/scss/utilities/mixins/cta/_cta--button.scss
yvonnetangsu: I didn't realize all image tags get this by default. Removed.

core/templates/components/cta/cta.twig
sherakama: These are great.

core/scss/utilities/mixins/cta/_cta--button.scss
sherakama: Ok great.

PR #253, opened 2018-11-01
make color change on hover & focus less abrupt

JBCSU:
While reviewing @sherakama's Card PR, I enjoyed the nice animation he added to the `action` link. Then I hovered over the card's button, and I found the color change very abrupt. I propose we add a little transition to buttons to make the color changes smoother.

### DIFF
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -1893,7 +1893,7 @@ button,
   text-align: center;
   text-decoration: none;
   width: 100%;
-  transition: all .25s ease-in-out; }
+  transition: background-color .25s ease-in-out, color .25s ease-in-out; }
   @media only screen and (min-width: 576px) {
     .su-button,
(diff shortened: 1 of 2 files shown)


core/css/decanter.css
```diff
@@ -1892,7 +1892,8 @@ button,
   outline: none;
   text-align: center;
   text-decoration: none;
-  width: 100%; }
+  width: 100%;
+  transition: all .25s ease-in-out; }
```
sherakama: Perhaps we could change 'all' to the specific transition as we have a known state. The `all` performance for the browser isn't as good as a specific transition. See: https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x

JBCSU: committed more specific targetting

PR #261, opened 2018-11-02
202 sticky footer

sherakama:
# READY FOR REVIEW

# Summary
- Creates a sticky footer mixin that puts the global footer component to the bottom of the page
- Adds the global footer to the rendered style guide
- I included the work from the homepage branch in here as well so you will see that code too.

# Needed By (Date)
- ?

# Urgency
- Highish as @JBCSU needs it next week.

# Steps to Test

0. Ignore the homepage.md and other related css gunk.
1. Check out this branch
2. Compile the styleguide
3. Browse the styleguide home page
4. See the global footer on the bottom
5. Open up your browser inspector 
6. Delete the `<article class="kss-main">` element from the dom to shorten the page's content
7. See the footer still on the bottom
8. Review code.

# See Also


core/scss/utilities/mixins/layout/_sticky-footer.scss
```diff
@@ -0,0 +1,27 @@
+@charset 'UTF-8';
+
+//
+// @sticky-footer
+//
+// Slam the footer to the bottom of the window below all content no matter
```
JBCSU:
```suggestion
// Stick the footer to the bottom of the browser window no matter
```

core/scss/utilities/mixins/layout/_sticky-footer.scss
```diff
@@ -0,0 +1,27 @@
...
+// the size of the browser or content.
+//
+// Apply this mixin to the body tag only. It will set the body tag height to
+// 100% and set it to display as #grid. This mixin will then target a direct
+// decendant element with the class `.global-footer` and set it to css grid's
```
JBCSU:
```suggestion
// descendant element with the class `.global-footer` and set it to css grid's
```

core/scss/utilities/mixins/layout/_sticky-footer.scss
```diff
@@ -0,0 +1,27 @@
...
+@mixin sticky-footer {
+  display: grid;
+  height: 100vh; // Avoid the IE 10-11 `min-height` bug.
+  @include padding(0);
+  @include margin(0);
+
+  > .global-footer {
+    align-self: flex-end;
+  }
+}
```
JBCSU: I think we should pass the selector for the footer as an argument to the mixin, rather than hardcoding it. On all of UComm's sites, including homesite, "global-footer" is just the universal footer; all our sites also have a fat footer above the universal footer, and both are wrapped in the `<footer>` element. We would want to replace `> .global-footer` with `> <footer>`. But I can see that wouldn't work for everyone.

core/scss/utilities/mixins/layout/_sticky-footer.scss
```diff
@@ -0,0 +1,27 @@
+@charset 'UTF-8';
+
+//
+// @sticky-footer
+//
+// Slam the footer to the bottom of the window below all content no matter
+// the size of the browser or content.
+//
+// Apply this mixin to the body tag only. It will set the body tag height to
```
JBCSU:
```suggestion
// Apply this mixin to the `<body>` tag only. It will set the body tag height to
```

Just a little pedantry. :)

core/scss/utilities/mixins/layout/_sticky-footer.scss
sherakama: I'd be ok with that. Keep the current model as the default but provide override.

JBCSU:
@sherakama I updated the `sticky-footer` mixin to take the footer's selector as a parameter. I updated the styleguide to try making its footer sticky to test it out. I'm not sure this approach is going to work. Download this branch, rebuild, then go to section 7, **Functions**. That section's short enough to see how the sticky footer behaves. It appears that, when a containing element has `display:grid`, all its direct descendants are given equal height. I couldn't find any value for `align-items` that would get them to just be their natural height.

I think this approach to sticking the footer to the bottom of the page will need more investigation.

sherakama:
@JBCSU I've updated the PR and have made a few adjustments but I am not sure that this is the best approach after all.

When setting the parent element (in this case body) to grid, all of the children elements are governed by CSS grid rules. Things like height, width, rows, spans, order, are all governed by grid. So in order to get this to work, all items that are directly below body will have to be 100% width and stack on one another. Nesting an item one layer deeper will allow for the natural flow of things again.

sherakama:
@JBCSU,

This is another pass at the sticky footer. I think this time it is much better. I went with flexbox instead of grid to accomplish this as there are an unknown number of items beneath the body and css grid likes known items.

Issue #222, opened 2018-10-24
Create a global header component (Masthead)


sherakama:
Create a global header component that combines the following other components

- Global Brand Bar
- Lockup and logo
- Primary Navigation
- Global search

**Note**: This exists in Redwood

buttonwillowsix:
This work is underway in Wordpress and might just need to be moved over.

yvonnetangsu:
done - .su-masthead

Issue #216, opened 2018-10-24
Create global search component


sherakama:
Create a global search element to match what is in the source Figma file.

PR #262, opened 2018-11-02
216 Create site search component

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Create the global site search component

# Needed By (Date)
- End of week

# Steps to Test

1. Pull this branch
1. Run grunt styleguide
2. Check that the search component matches the design in the Figma file "Stanford Design Library" -> "Header, Footer & Navigation" section. Check that the twig file is properly set up and that basic accessibility requirements are met.

- Decanter

- https://github.com/SU-SWS/decanter/issues/216
- https://github.com/SU-SWS/decanter/issues/222
- For xs to md break points according to Figma documentation, this component lives within the main nav component, and is hidden from view until the main nav is expanded.


core/templates/components/site-search/site-search.twig
```diff
@@ -0,0 +1,26 @@
...
+ * - input_attributes: Additional attributes for the input field
+ * - placeholder: Placeholder text for the input field
+ * - button_attributes: Additional button attributes
+ * - additional_fields: Any additional fields, e.g., hidden input fields
+ *
+ */
+#}
+<div class="su-site-search" role="search">
+ <form action="{{ action }}" method="{{ method }}" accept-charset="UTF-8">
+  <label class="su-site-search__sr-label" for="search-field">Search this site</label>
```
JBCSU:
```suggestion
  <label class="su-site-search__sr-label" for="search-field">{{ placeholder }}</label>
```

core/templates/components/site-search/site-search.twig
```diff
@@ -0,0 +1,26 @@
...
+  <input{{ input_attributes }} type="text" id="search-field" class="su-site-search__input" placeholder="{{ placeholder }}" maxlength="128">
+  <button{{ button_attributes }} type="submit" class="su-site-search__submit" aria-label="Search"><img src="../../../img/icon-search.svg" alt="Search button"/></button>
```
JBCSU: How are we handling paths to images? Is it okay to use relative paths within Decanter?

Also, I can imagine some sites might want to use a different icon, or have no icon. Should the icon be a parameter, and / or optional?

core/scss/components/site-search/_site-search.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  &:active,
+  &:focus {
+    box-shadow: none;
+  }
+
+  &:focus {
+    outline: #aaa solid 3px;
+    outline: -webkit-focus-ring-color auto 5px;
```
JBCSU: Why are you changing the focus behavior just on webkit browsers? Should we specify this universally and let autoprefixer manage the vender prefix?

core/scss/components/site-search/_site-search.scss
```diff
@@ -0,0 +1,73 @@
...
+  border-radius: 20px;
+  font-size: .8em;
+
+  &::placeholder {
+    color: $color-black;
+    opacity: 1; // For Firefox
+  }
+
+  @media #{$breakpoint-lg} {
+    width: 320px;
```
JBCSU: Would `max-width` be better here?
Would the value be better specified in (r)em's?

core/scss/components/site-search/_site-search.scss
```diff
@@ -0,0 +1,73 @@
...
+// Markup: ../templates/components/site-search/site-search.twig
+//
+// Style guide: Components.Site Search
+//
+.su-site-search__input {
+  @include font-smoothing();
+  @include padding(.3em 1em);
+
+  display: inline-block;
+  height: 40px;
```
JBCSU: Would min-height be better here?
Would the value be better specified in (r)em's?

core/templates/components/site-search/site-search.twig
yvonnetangsu: I went for a more hardcoded approach since accessibility for search takes a little research to do right so I thought perhaps this would prevent people from making it not accessible. I'm open to making some of these elements into variables if needed.

yvonnetangsu: @JBCSU, I didn't use the placeholder variable for the label field since the word "search" is required for accessibility for the label, and we don't know if people will use the word "search" for the placeholder. Let me know what you think.

core/scss/components/site-search/_site-search.scss
yvonnetangsu: It's just that webkit has their own default focus outline setting (the light fuzzy light blue) so I want it to be consistent with other elements on focus when they're using a webkit browser. There's no similar setting for other browsers. I don't believe autoprefixer deals with that particular item.

core/templates/components/site-search/site-search.twig
yvonnetangsu: Hmm....this might get a little tricky since I'm using the icon as the submit button here and a submit button is recommended for accessibility. Or maybe we can do a variant with a text button or no button (is this ok - might need to ask Jiatyan)?

yvonnetangsu: The path and image is a great question, and I think I'll make the img src to be a variable so people have the option to use a different icon at least. Thanks.

core/scss/components/site-search/_site-search.scss
yvonnetangsu: Yes, I'm changing it to max-width and using 20em.

core/scss/components/site-search/_site-search.scss
yvonnetangsu: I'm leaving the height as px for now, since in Figma the height of the bar stays the same for all break points, and using em might change the height for different break points. I'm not super up-to-date with the typography settings - is rem always going to be 10px?

core/templates/components/site-search/site-search.twig
```diff
@@ -0,0 +1,27 @@
...
+ * - placeholder: Placeholder text for the input field
+ * - button_attributes: Additional button attributes
+ * - search_icon_src: The href property of the icon image
+ * - additional_fields: Any additional fields, e.g., hidden input fields
+ *
+ */
+#}
+<div class="su-site-search" role="search">
+ <form action="{{ action }}" method="{{ method }}" accept-charset="UTF-8">
+  <label class="su-site-search__sr-label" for="search-field">Search this site</label>
```
sherakama:
```suggestion
  <label class="su-site-search__sr-label" for="search-field">{{ search_label|default('Search this site') }}</label>
```

sherakama: I like the option of being able to change this field as well. Not allowing this will likely lead to someone doing wacky things with Twig includes or copy and paste with html.

core/templates/components/site-search/site-search.twig
```diff
@@ -0,0 +1,27 @@
...
+  <input{{ input_attributes }} type="text" id="search-field" class="su-site-search__input" placeholder="{{ placeholder }}" maxlength="128">
```
sherakama: field name?

core/templates/components/site-search/site-search.twig
```diff
@@ -0,0 +1,27 @@
...
+ * - method: Method of the form field
+ * - input_attributes: Additional attributes for the input field
+ * - placeholder: Placeholder text for the input field
+ * - button_attributes: Additional button attributes
+ * - search_icon_src: The href property of the icon image
+ * - additional_fields: Any additional fields, e.g., hidden input fields
+ *
+ */
+#}
+<div class="su-site-search" role="search">
```
sherakama: missing attributes and modifier class variables.

core/templates/components/site-search/site-search.twig
yvonnetangsu: That is true. Applied your suggestion and will make other changes soon. Thanks!

core/templates/components/site-search/site-search.twig
yvonnetangsu: @JBCSU, I've applied Shea's suggestion and made the search label a variable. He too thinks that people might want to customize this field and not letting them do so might lead to them using "unconventional" solutions :)

core/templates/components/site-search/site-search.twig
yvonnetangsu: Added

core/templates/components/site-search/site-search.twig
yvonnetangsu: Added - using same variable for search input name and ID, and also its label's "for" attribute. Let me know if you think that's a problem for them to be all the same.

Issue #189, opened 2018-08-15
Update Alerts molecule


josephgknox:
- John H will style and add wireframe mockup

josephgknox:
We do not have a wireframe or mock up for this, so will clean up what is there.

PR #263, opened 2018-11-02
189 Alert Component

josephgknox:
# READY FOR REVIEW

# Summary
- Updated Alert component

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

2. Review visual of alert component and variants
3. Review all related code

- Decanter

- #189 

# See Also


core/templates/components/alert/alert.twig
```diff
@@ -1,4 +1,23 @@
...
+ *   - Default: Black text with gray background with no icon
+ *   - Success: Black text with green background with check icon
+ *   - Warning: Black text with yellow background with traiangular exclamation mark icon
+ *   - Error: Black text with red background with circlular exclamation mark icon
+ *   - Info: Black text with blue background with information icon
+ *
+ * Available variables:
+ * - attributes: For additional HTML attributes not already provided.
+ * - modifier_class: Additional css classes to change look and behaviour.
+ */
```
sherakama: Missing header and text variables.

core/scss/components/alert/_alert.scss
```diff
@@ -7,18 +7,56 @@
...
+// .su-alert--info     - Informative Message
 //
 // Markup: ../templates/components/alert/alert.twig
 //
 // Style guide: Components.Alert
 //
 .su-alert {
-  @include alerts;
-}
+  @include margin(1.5em null null);
```
yvonnetangsu: I only have one question - is there a reason to add a top margin for this component? We don't have margins for cards or hero block for example. Everything else looks great to me!

sherakama: Oh good catch!

@josephgknox What gives?

josephgknox: Just keeping you both on your toes. :)

Really, this shouldn't be there, as this type of spacing/margin should be applied more globally (theme level, project level, etc.) rather than on an individual component.

Thanks for catching this. I've pushed up a fix.

PR #284, opened 2019-01-21
Add region attribute variables to grid templates.

sherakama:
# Ready for review

# Summary
- Adds optional region wrapper attributes. 
- For integration into CMS systems that require markup be injected here
- Creates a standard in which to apply to components for nested region attributes

While working on the Drupal integration I came across a need to inject attributes around each of the variable regions. This is the standard they use. They use something called region_attributes and nest all of their attributes in that variable as an object. This allows me to add css classes, data attributes, ids, and more to the wrappers but does not introduce any extra markup for those that don't need them. We have come across the need for this kind of approach in the components with some nested attributes and would like to refactor to match.

### DIFF
--- a/core/templates/layout/grid/four-column/molive/molive.twig
+++ b/core/templates/layout/grid/four-column/molive/molive.twig
@@ -22,7 +22,7 @@
     {# A centered max-width header region #}
     {% if header is not empty %}
     <header>
-      <div>
+      <div {{ region_attributes.header }}>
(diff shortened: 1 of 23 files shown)


josephgknox:
@yvonnetangsu is this approach OK with you? If so, I'll merge.

yvonnetangsu:
Yes, sounds good to me! Thanks!

PR #295, opened 2019-02-21
Set base font size for each breakpoint

yvonnetangsu:
# READY FOR REVIEW

# Summary
- A small PR - Just setting the base font size at each breakpoint as described in Figma doc so we can use em values properly in other places.

# Needed By (Date)
- N/A

# Urgency
- Not urgent

# Steps to Test

2. Check that `<body>` has the correct base font size at each breakpoint.

- Decanter

- Might affect other components/elements and the styleguide itself that uses "em" values.
- JB is still working on the other modular-typography related mixins (eg, display-a) so those will come later.


yvonnetangsu:
screen shot 2019-02-21 at 3 26 18 pm

PR #297, opened 2019-02-22
72 main nav multiple levels

sherakama:
Refactors menu loop to support infinite levels.

Needs some css and js work to get the third level right.

### DIFF
--- a/core/templates/components/main-nav/main-nav.json
+++ b/core/templates/components/main-nav/main-nav.json
@@ -10,16 +10,46 @@
       "href": "#",
       "text": "Useful Websites",
       "current": true,
-      "lv2": [
+      "children": [
         {
           "href": "https://www.stanford.edu",
           "text": "Stanford",
-          "target": "_blank"
+          "attributes": {
+            "target": "_blank"
+          }
         },
         {
           "href": "https://uit.stanford.edu/sws",
           "text": "Stanford Web Services",
-          "target": "_blank"
+          "attributes": {
+            "target": "_blank",
+            "rel": "external"
+          },
+          "children": [
+            {
+              "href": "https://www.stanford.edu",
+              "text": "Stanford",
+              "attributes": {
+                "target": "_blank"
+              }
+            },
+            {
+              "href": "https://uit.stanford.edu/sws",
+              "text": "Stanford Web Services",
+              "attributes": {
+                "target": "_blank"
+              }
+            },
+            {
+              "href": "https://ucomm.stanford.edu/",
+              "text": "University Communications"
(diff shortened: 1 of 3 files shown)


yvonnetangsu:
Cool - thanks for working on this. Will take a look when I have a chance and see what needs to be done.

sherakama:
Just FYI, I had to use a version of this in our Drupal theme to get it to work.

yvonnetangsu:
Hmm....I wonder why 🤔 . Scott was able to use the one on master in the event calendar site. I pulled down this branch but haven't had a chance to work on it yet. I suspect the js would be the tricky part which probably JB would be handling. I know she's making great progress on webpack so let's wait till that's done 😄 ?

sherakama:
It more has to do with the format of information that is available to pass through into the structured output.  You can see how I patched through the `{{ items }}` array that has the menu information in to the macro loop.

here:
https://github.com/SU-SWS/stanford_basic/blob/8.x-4.x/templates/menus/menu--main.html.twig

and here:
https://github.com/SU-SWS/stanford_basic/blob/8.x-4.x/templates/menus/macros/main-nav-menu.twig

JBCSU:
> Hmm....I wonder why 🤔 . Scott was able to use the one on master in the event calendar site.

[Scott's nav](http://events-dev.stanford.edu/) actually only appears to have 3 levels. In fact it's 2 levels, and he (we) added a styling override to make a few items appears indented relative to **All Arts** in the **By Subject** dropdown.

yvonnetangsu:
> > Hmm....I wonder why 🤔 . Scott was able to use the one on master in the event calendar site.
> 
> [Scott's nav](http://events-dev.stanford.edu/) actually only appears to have 3 levels. In fact it's 2 levels, and he (we) added a styling override to make a few items appears indented relative to **All Arts** in the **By Subject** dropdown.

Oh, I actually meant the regular 2 level nav. @sherakama , when you said "I had to use a version of this in our Drupal theme to get it to work" do you mean 2 level or 3 level? I interpreted that as you had to tweak the twig template if you want the main nav to work in Drupal.

sherakama:
I only have the two levels working in Drupal as that is all the styles support at the moment. 

Drupal doesn't send me data with items keyed the same way as in the .json example. This is to be expected in any system so I have to make a choice. Try to reformat the data Drupal sends to match the example data or reformat the loop. Drupal sends me well structured data with child/parent relationships and attributes that nest several levels deep so it was an easy choice for me.

sherakama:
This is too far behind now. Going to refactor and open up a new one.

Issue #298, opened 2019-02-25
target @mixin sticky-footer to .site-footer selector instead of .global-footer; change su-global-footer component to <section> rather than <footer>


stocker:
**Is your feature request related to a problem? Please describe.**
@mixin sticky-footer targets .global-footer, which means that only the red global footer gets pinned to the bottom of the window.  Targeting the <footer> element instead means that the entire footer, including site-specific footer, will get pinned to bottom of window.

**Describe the solution you'd like**
Change @mixin sticky-footer to target the <footer> selector.

**Describe alternatives you've considered**
n/a

**Additional context**

sherakama:
Targeting an HTML tag would be difficult. We would not want to affect all `<footer>` elements but rather only the one at the bottom of the page. We also, cannot assume where in the DOM this element will be. It could be a direct child of `<body>` or it could be a number of children deeper. 

Perhaps we flip this problem on its head and make the sticky footer its own CSS class that the site builder can add to their site and add a variant for the global footer so that the default is not sticky.

stocker:
good thought.  Should I adjust to <footer class="sticky-footer"> and target the mixin to .sticky-footer?  I may also need some guidance on how to commit to a pending pull-request.

stocker:
changed sticky-footer mixin to target ".page-footer" class by default rather than  <footer> element.  selector can be overridden as well.

stocker:
these changes have been merged into master.  closing this issue.

PR #299, opened 2019-02-25
298-change "sticky-footer" mixin to target <footer> selector 

stocker:
change "sticky-footer" mixin to target <footer> selector  rather than .global-footer.  In global-footer.twig, move .global-footer class to a section inside of footer instead of adding to <footer> element.  This allows addition of a .site-footer section to <footer>, which will also stick to bottom of window.

### DIFF
--- a/core/scss/components/brand-bar/_brand-bar.scss
+++ b/core/scss/components/brand-bar/_brand-bar.scss
@@ -21,9 +21,6 @@
 
 .su-brand-bar__container {
   @include centered-column;
-  @include grid-media-only('xs') {
-    @include padding(null 20px);
-  }
 }
 
 .su-brand-bar__logo {
--- a/core/templates/components/global-footer/global-footer.twig
+++ b/core/templates/components/global-footer/global-footer.twig
@@ -1,18 +1,18 @@
-<footer{{ attributes }}>
-  <section class="global-footer {{ modifier_class }}">
-    <div class="global-footer__container">
-      <div class="global-footer__brand">
-        <a href="https://www.stanford.edu/" class="global-footer__logo">Stanford<br>University</a>
+<footer{{ attributes }} class="page-footer  {{ modifier_class }}">
+  <section class="su-global-footer">
+    <div class="su-global-footer__container">
+      <div class="su-global-footer__brand">
+        <a href="https://www.stanford.edu/" class="su-global-footer__logo">Stanford<br>University</a>
       </div>
-      <div class="global-footer__content">
+      <div class="su-global-footer__content">
         <nav role="navigation">
-          <ul class="global-footer__menu global-footer__menu--global">
+          <ul class="su-global-footer__menu su-global-footer__menu--global">
             <li><a href="https://www.stanford.edu">Stanford Home</a></li>
             <li><a href="https://visit.stanford.edu/plan/">Maps &amp; Directions</a></li>
             <li><a href="https://www.stanford.edu/search/">Search Stanford</a></li>
             <li><a href="https://emergency.stanford.edu">Emergency Info</a></li>
           </ul>
-          <ul class="global-footer__menu global-footer__menu--policy">
+          <ul class="su-global-footer__menu su-global-footer__menu--policy">
             <li><a href="https://www.stanford.edu/site/terms/" title="Terms of use for sites">Terms of Use</a></li>
             <li><a href="https://www.stanford.edu/site/privacy/" title="Privacy and cookie policy">Privacy</a></li>
             <li><a href="https://uit.stanford.edu/security/copyright-infringement"
@@ -25,11 +25,11 @@
             </li>
           </ul>
         </nav>
-        <div class="global-footer__copyright">
+        <div class="su-global-footer__copyright">
           <span>&copy; Stanford University.</span>
           <span>&nbsp; Stanford, California 94305.</span>
         </div>
       </div>
     </div>
-  </section>
-</footer> <!-- global-footer end -->
+  </section> <!-- su-global-footer end -->
+</footer>
--- a/core/templates/components/global-footer/global-footer.twig
+++ b/core/templates/components/global-footer/global-footer.twig
@@ -1,35 +1,45 @@
-<footer{{ attributes }} class="page-footer  {{ modifier_class }}">
-  <section class="su-global-footer">
+{#
+/**
+ * @file
+ * Global Footer Component.
+ *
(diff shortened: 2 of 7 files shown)


kss/builder/decanter/kss-assets/css/kss.css
```diff
@@ -867,7 +867,7 @@ body {
   -webkit-box-direction: normal;
       -ms-flex-direction: column;
           flex-direction: column; }
-  body > .global-footer {
+  body > footer {
```
sherakama: We cannot assume that the element we want to target will be a direct descendant of the body tag. See comments in https://github.com/SU-SWS/decanter/issues/298

core/templates/components/global-footer/global-footer.twig
```diff
@@ -1,29 +1,35 @@
...
-          <li><a href="https://uit.stanford.edu/security/copyright-infringement" title="Report alleged copyright infringement">Copyright</a></li>
-          <li><a href="https://adminguide.stanford.edu/chapter-1/subchapter-5/policy-1-5-4" title="Ownership and use of Stanford trademarks and images">Trademarks</a></li>
-          <li><a href="http://exploredegrees.stanford.edu/nonacademicregulations/nondiscrimination/" title="Non-discrimination policy">Non-Discrimination</a></li>
-          <li><a href="https://www.stanford.edu/site/accessibility" title="Report web accessibility issues">Accessibility</a></li>
-        </ul>
-      </nav>
-      <div class="global-footer__copyright">
-        <span>&copy; Stanford University.</span>
-        <span>&nbsp; Stanford, California 94305.</span>
+<footer{{ attributes }} class="page-footer  {{ modifier_class }}">
```
sherakama: Unfortunately, this breaks our BEM naming conventions. The outer most wrapper must be the block for every component and the nested children elements follow the block__element naming convention. This change does bring up the slight inconsistency in our naming pattern with the original block level class name in that it should have been `su-global-footer` to have started with.

josephgknox: So, would grandchild selectors be our friend here? Something like:

```
<footer{{ attributes }} class="su-page-footer  {{ modifier_class }}">
  <section class="su-page-footer__global-footer">
    <div class="su-page-footer__global-footer__container">
      ...
```

Would this work? Do we only want underscores to appear once in a selector name? I, of course, we should avoid too often ending up with a component structure that leads to something like `parent__child__grandchild__greatgrandchild`, but perhaps this is an example where the grandchild depth is OK.

BEM naming isn’t strictly tied to the DOM and doesn’t limit the nested levels of a descendent element, but is rather available to help us identify the relationship with the top level of a component,  in our case, the page footer. That said, the above might work (?).

Whatcha think? Am I crazy?

sherakama: To be honest, I think we instead of refactoring the component, we simply remove the stickiness from its default state and make it a variant instead. This would keep the component consistent to where it is now and allow the implementor to add a css class to whatever element they want 'stuck' to the bottom. See https://github.com/SU-SWS/decanter/pull/300 for example.

stocker:
The issue I encountered that lead us down this detour is that my event calendar use case includes both a fat-footer and the red global footer (similar layout to on the SU Homepage).   The existing implementation applied the .global-footer class to the <footer> element (ie <footer class=”global-footer”>.)  That applies the red background color to the entire <footer>, making it challenging to include the fat-footer region which isn’t part of that component.  I refactored my code for the events page footer to be this structure:

<footer>
<section id=”site-footer”><!—fat footer--></section>
<section id=”global-footer” class=”global-footer”><!—global footer--></section>
</footer>

The global footer component is only the red bar, and I think should be defined as a section of <footer> rather than the entire <footer>.

I personally like the sticky-footer mixin because it can be included and target an element or section without having to use an exact class name like “global-footer—sticky” in your markup.  Now that I understand how you can override the selector in the mixin, I’m actually OK with just going back to the original implementation, even though I think it may be a bit confusing to suggest in our twig template that the entire <footer> element needs to have the .global-footer class applied to it when it is fairly common to have the fat-footer in additional to the Global Footer.

Hope that makes sense.

Scott

yvonnetangsu:
I'm also liking the idea of the global-footer being contained within a `<section>` tag instead since that allows for more flexibility. Also, the brand bar is contained within a `<section>` tag as well so they can mirror each other that way.

sherakama:
Ah, I failed to grok that and can buy in. 
Instead of adding another level, how about just changing the outer most wrapper of the global footer so that it is not a `<footer>` tag but rather a `<section>` tag. Unless we add a region to the global footer (which I don't think we should do) I'd rather see this component used "as is" and placed inside a site's custom work rather than modified afterward.

stocker:
This sounds like a good plan to me.  Let me make one more pass in the pull request that I had created, and I’ll try to get it ready for review in advance of the working session meeting tomorrow.  Thanks for all the good input on this.  I think we’ve nailed it down now!

Scott

yvonnetangsu:
@stocker @sherakama 
I added a few really minor tweaks. Let me know if they're ok. I'd be happy to reverse any if there are issues.

1. Matched the indenting of global-footer.twig to the style of the other twig files (1 space per tab instead of 2 spaces per tab).
2. Changed the default selector for sticky-footer to `.site-footer `instead of `.page-footer`. 
3. Cleaned up some documentation (eg, there was a line "This file is here until https://github.com/SU-SWS/decanter/pull/261 gets merged...." so that was removed.)
4. Removed the left/right padding for brand bar at XS breakpoint. We now have the screen edges merged in so we don't need the padding anymore.

josephgknox:
@yvonnetangsu :)
tabs-vs-spaces-640x360

yvonnetangsu:
@josephgknox Obviously it's top priority 😼

core/scss/components/brand-bar/_brand-bar.scss
```diff
@@ -21,9 +21,6 @@
 
 .su-brand-bar__container {
   @include centered-column;
-  @include grid-media-only('xs') {
```
sherakama: Good catch. Thanks.

kss/builder/decanter/index.twig
```diff
@@ -273,6 +273,8 @@
   });
 </script>
 {{ scripts|raw }}
+<footer class="page-footer">
```
yvonnetangsu:
```suggestion
<footer class="site-footer">
```
I forgot to change the class name for the kss twig after I changed it in the mixin.

Issue #120, opened 2018-05-23
Skip links


sherakama:
Skip links require:
• the link(s) themselves - we put them in the `<header>`, above the brand bar, on [homesite](https://www.stanford.edu/);
• correct styling for screen-reader text - I don’t think what’s currently in Decanter is what WCAG recommends - see [homesite](https://www.stanford.edu/) for our SR styling; and
• a target for the link(s) - we found that the best target was a screen-reader-only `<h2>` with the appropriate text, e.g. “Main Content”.

sherakama:
Closing in favor of the PR.

PR #303, opened 2019-02-28
120 Skiplinks

josephgknox:
# READY FOR REVIEW

# Summary
- Adds Skiplinks component and class
- Refactors screen reader only class(es), leveraging some Bourbon mixins
- General related cleanup

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Run `grunt styleguide`
3. Review Skiplinks component

Note: the `z-index` of the brand bar and the Decanter header in the KSS style guide overlay the "Skip to content" button link that slides into view on :focus. Those indexes might need to be temporarily disabled (in inspector) to test this fully.

- Decanter

- #120 

# See Also


josephgknox:
Questions:

- Homesite uses an ID rather than a class for targeting skiplinks. Do we want to do the same? That is, `#su-skiplinks` rather than `.su-skiplinks`?
- Do we want to assign a `z-index` to the Skiplinks component when it slides into view?
- If and when this is merged in, do we want to drop our current Skipnav component and related code?

core/templates/components/skiplinks/skiplinks.twig
```diff
@@ -0,0 +1,7 @@
+<header>
+  <a href="#main-content" class="su-skiplinks">{{ text|default('Skip to content') }}</a>
```
sherakama: Id's are hard and don't really belong in a pattern library. There is no way to ensure that the next developer's software doesn't or does have a main-content id. Unfortunately, for this functionality, we need one. I would suggest opening up the anchor link to a twig variable and perhaps generate a random id if one is not provided.

Example not tested*
```
{% set skipID = "skip-to-content-" ~ random() %
 <a href="{{ anchor|default(skipID) }}" class="su-skiplinks">{{ text|default('Skip to content') }}</a>
```

sherakama:
> Do we want to assign a `z-index` to the Skiplinks component when it slides into view?

I am not sure. I wouldn't want it to be behind anything so, yes? As this is likely above the brand bar perhaps something higher than that component?

> If and when this is merged in, do we want to drop our current Skipnav component and related code?
Yes, out with the old, in with the new.

yvonnetangsu:
> * Homesite uses an ID rather than a class for targeting skiplinks. Do we want to do the same? That is, `#su-skiplinks` rather than `.su-skiplinks`?

For the skiplink component itself, I'm totally fine with just using a class .su-skiplink. I believe some sites have multiple skiplinks, eg, one link to content and one to footer, then in that case, it makes even more sense to just use a class.

> * Do we want to assign a `z-index` to the Skiplinks component when it slides into view?

Like @sherakama , I do like the skiplinks to be above everything else when it comes into view. I think somewhere in my nav I set something to z-index: 22222 so any number larger than that is probably good 😂 

> * If and when this is merged in, do we want to drop our current Skipnav component and related code?

Yes, I'm totally ok with having just the new skiplink component and drop the old one. It would be nice to be able to pass href variable though.

I do have a question about the twig template. Is it more for demo purposes rather than something that developer will actually pull into their site template? I imagine that the skiplink mixin is easier to use since it doesn't add the header and main sections.

Thanks for working on this, @josephgknox! This is so much better than the old one 👍

josephgknox:
Thank you both for the feedback. I will continue working to implement some of the above, as well as clean up the old Skipnav component.

@yvonnetangsu good question. More so for demo purposes, as I wanted it to be testable in some way via the pattern library. I, too, think that a developer would use the mixin/class/ID/etc. as they wish, leveraging the functionality more than the template.

sherakama:
@josephgknox, for the component itself, have a usable template that could be embedded directly. If you need to show it as an example in the styleguide, create a new template or use straight up markup to change the rendered output example.

yvonnetangsu:
@josephgknox - Great! That makes sense. In that case, would it be helpful to add some wording to explain how one can test the demo in the component section since it's not visible until they start tabbing? Before I started working here, I wasn't familiar with skiplinks and wouldn't know to use the tab key 😂 so perhaps a short explanation would help developers who are new to this.

josephgknox:
@sherakama @yvonnetangsu can you two have another review? We closer?

core/templates/components/skiplinks/skiplinks.twig
```diff
@@ -0,0 +1,9 @@
+{% set skipID = "skip-to-content-" ~ random() %}
+
+<header>
+  <a href="#{{ anchor|default(skipID) }}" class="su-skiplinks">{{ text|default('Skip to content') }}</a>
+</header>
+
+<main>
+  <h2 id="{{ skipID }}" class="su-sr-only-text" tabindex="-1">{{ skipped_to_title }}</h2>
```
JBCSU: The id of the target element should match the href of the skip link, i.e. `#{{ anchor|default(skipID) }}`.

Or `#{{ target_id|default(skipID) }}` if you accept my suggestion on the `.json` file.

core/templates/components/skiplinks/skiplinks.json
```diff
@@ -0,0 +1,3 @@
+{
+  "skipped_to_title": "Main Content"
```
JBCSU: I would do:
```
{
  target_id: "main-content",
  target_text: "Main Content",
  link_text: "Skip to main content
}
```
and make the corresponding changes in the `.twig` file.

core/scss/utilities/mixins/accessibility/_skiplinks.scss
```diff
@@ -0,0 +1,57 @@
...
+    border: 1px solid $color-cool-grey;
+    border-radius: $border-radius;
+    height: auto;
+    left: .8em;
+    position: fixed;
+    top: .8em;
+    width: auto;
+    z-index: 9999;
+  }
+}
```
JBCSU: Why doesn't this extend the helper classes, `.su-sr-only-element` or `.su-sr-only-text`, or include the mixins, `hide-visually`, `hide-text`?

core/scss/elements/input/_input.scss
```diff
@@ -74,7 +74,7 @@ fieldset {
 [type=radio] {
   // The actual input element is only visible to screen readers, because
   // all visual styling is done via the label.
-  @include sr-only;
+  @include hide-visually;
```
JBCSU: Really? All checkboxes and radio buttons are hidden by default?

yvonnetangsu:
> @sherakama @yvonnetangsu can you two have another review? We closer?

Yes, definitely closer! A few suggestions - please feel free to push back.
1. Could we give it a z-index of something like 11222 since I use 11111 for the main nav submenu and it might be better to make sure the skiplink is always above any open submenus?
2. `#kssref-components-skiplinks .kss-modifier__example {z-index: 11222;} `
might help bring it above the brand bar when it's on focus. I removed the global z-index on the .kss-modifier__example containersa while back since it really messed with the main nav z-index.
3. Could we add something to the doc along the line of "To see the skiplink component in action, click to the space to the right of the word "Example" in the below container and hit the "tab" key"? (feel free to modify) 😄

JBCSU:
> 1. Could we give it a z-index of something like 11222 since I use 11111 for the main nav submenu and it might be better to make sure the skiplink is always above any open submenus?

For identity and homesite, Kevin set up a single separate file that manages all explicit z-indexes for All The Things. I've found it *extremely* helpful to be able to go to just one place and see the relative "heights" of various elements. I heartily suggest doing that in Decanter. I think it would make it much easier for people building on top of Decanter to find the stack order in a single location so they could know definitely where they need to stick their stuff to show up at the right level.

yvonnetangsu:
> > 1. Could we give it a z-index of something like 11222 since I use 11111 for the main nav submenu and it might be better to make sure the skiplink is always above any open submenus?
> 
> For identity and homesite, Kevin set up a single separate file that manages all explicit z-indexes for All The Things. I've found it **extremely** helpful to be able to go to just one place and see the relative "heights" of various elements. I heartily suggest doing that in Decanter. I think it would make it much easier for people building on top of Decanter to find the stack order in a single location so they could know definitely where they need to stick their stuff to show up at the right level.

That does sound like a good idea. I wonder what's the best way to do this. Maybe have the z-indexes in a variable map and pull the value for each component as needed?

core/templates/components/skiplinks/skiplinks.twig
josephgknox: Added!

core/templates/components/skiplinks/skiplinks.json
josephgknox: Added!

josephgknox:
@yvonnetangsu thanks! I will update the default `z-index` and will add a partial to the KSS SCSS to target the style guide itself. Also, +1 to some sort of documentation around this. Once things are more "approved" as far as syntax, organization, etc. goes, let's make sure we capture something.

core/scss/elements/input/_input.scss
josephgknox: This component has not been touched for a very long time, and includes (mostly) code from the U.S. Web Design System (https://github.com/uswds/uswds). Currently, the visual styling for these are being done on the label, which is what the input element itself is "hidden." Nonetheless, I think we'll want to revisit this and make it mo betta overall.

Because I dropped the `sr-only` mixin in favor of the`hide-visually` and `hide-text` Bourbon mixins, the compiler was throwing an error ("...cannot find `sr-only` mixin..."). So, I quickly replaced it, which is why it is appearing as part of this PR.

josephgknox:
> > 1. Could we give it a z-index of something like 11222 since I use 11111 for the main nav submenu and it might be better to make sure the skiplink is always above any open submenus?
> 
> For identity and homesite, Kevin set up a single separate file that manages all explicit z-indexes for All The Things. I've found it **extremely** helpful to be able to go to just one place and see the relative "heights" of various elements. I heartily suggest doing that in Decanter. I think it would make it much easier for people building on top of Decanter to find the stack order in a single location so they could know definitely where they need to stick their stuff to show up at the right level.

I also like this idea! Do we want to break this out into another task? Or, do we want to complete something as part of this task/PR?

yvonnetangsu:
> I also like this idea! Do we want to break this out into another task? Or, do we want to complete something as part of this task/PR?

Thanks for the fixes! And I vote for breaking this into another task so we can merge this in faster.

core/scss/utilities/mixins/accessibility/_skiplinks.scss
josephgknox: It can. The approach here, though, is a little bit different than that of what those two mixins (or related classes) do. I used Homesite as an example of an approach and aimed to mirror that implementation, which only "hides" something from print, but does not leverage nor extend the Bourbon mixins (hiding the element or the text).

While `@include hide-visually;` and `@include hide-visually("unhide");` for hover and focus states might get us something similar as far as behavior, it would be a slight shift from what Homesite is doing, I believe.

Related: I noticed that Homesite had a `_placeholders.scss` file that has a primary purpose of being available for extended code. If we do want to use extends here (or in the future), we might want to consider doing something similar.
https://code.stanford.edu/ucomm/homesite-2017/blob/master/wp-content/themes/homesite17/scss/base/_placeholders.scss

JBCSU:
> > > 1. Could we give it a z-index of something like 11222 since I use 11111 for the main nav submenu and it might be better to make sure the skiplink is always above any open submenus?
> > 
> > 
> > For identity and homesite, Kevin set up a single separate file that manages all explicit z-indexes for All The Things. I've found it **extremely** helpful to be able to go to just one place and see the relative "heights" of various elements. I heartily suggest doing that in Decanter. I think it would make it much easier for people building on top of Decanter to find the stack order in a single location so they could know definitely where they need to stick their stuff to show up at the right level.
> 
> I also like this idea! Do we want to break this out into another task? Or, do we want to complete something as part of this task/PR?

I've created issue #305 to capture this task. Let's not let scope-creep delay merging the good work you've already done.

core/scss/elements/input/_input.scss
JBCSU: Roger that. Thanks for the explanation. I do think we'll want to revisit this, but not as part of this PR.

core/templates/components/skiplinks/skiplinks.twig
JBCSU: Thank you!

core/templates/components/skiplinks/skiplinks.json
JBCSU: Thank you!

core/scss/utilities/mixins/accessibility/_skiplinks.scss
JBCSU: Do you know specifically what would be different if we used `hide-visually`? Is it the animation of having the skip links drop down from `top:-500px`?

I'm ok with this as is. I was just curious why we're not using the utilities that were updated as part of this PR.

core/templates/components/skiplinks/skiplinks.twig
```diff
@@ -0,0 +1,9 @@
+{% set skipID = "skip-to-content-" ~ random() %}
+
+<header>
```
sherakama: Do we need a wrapper around this? Isn't the idea that this link be placed as the first element after the body tag? If that is so, I would recommend removing or changing the `<header>` tag as to not compete with other header elements right after the body tag.

core/templates/components/skiplinks/skiplinks.twig
```diff
@@ -0,0 +1,9 @@
...
+  <a href="#{{ target_id|default(skipID) }}" class="su-skiplinks">{{ link_text|default('Skip to main content') }}</a>
```
sherakama: Missing modifier class.

sherakama: Missing attributes

core/templates/components/skiplinks/skiplinks.twig
```diff
@@ -0,0 +1,9 @@
...
+</header>
+
+<main>
```
sherakama: Do we need to have the target in this template file? It won't be directly usable if this main markup is here. I am assuming this is an example. I would like to see a directly usable option as well as the demo.

core/templates/components/skiplinks/skiplinks.twig
josephgknox: Thanks - this has been removed.

core/templates/components/skiplinks/skiplinks.twig
josephgknox: @sherakama is your recommendation that the twig template only include the following?

```
{% set skipID = "skip-to-content-" ~ random() %}

<a{{ attributes }} href="#{{ target_id|default(skipID) }}" class="su-skiplinks {{ modifier_class }}">{{ link_text|default('Skip to main content') }}</a>
```

And, then, perhaps add exampel markup directly to the SCSS file, something like:

```
<a href="#main-content" class="su-skiplinks ">Skip to main content</a><main><h2 id="main-content" class="su-sr-only-text" tabindex="-1">Main Content</h2></main>
```

JBCSU: I think skiplinks are a special snowflake. They are, by definition, not self-contained. I don't believe there's anyway to get them to work with a single twig file.

Since they'll require the user to create custom markup anyway, I vote for having all the necessary pieces in our `.twig` file, and document how users need to use each piece. I would therefore vote for restoring the `<header>` wrapper around the link itself, since it demonstrates the recommended placement for the link.

josephgknox:
@yvonnetangsu @JBCSU @sherakama 

I think this is now ready for a final review and/or further discussion on any of the recent comments and changes.

core/scss/components/skiplinks/_skiplinks.scss
```diff
@@ -0,0 +1,30 @@
...
+// Skiplinks
+//
+// For providing a link at the top of the page which jumps
+// the user down to an anchor or target at the beginning of the main content.
+//
+// Gives screen readers and keyboard users the same capability of going directly
+// to the main content as sighted, mouse users.
+// of the main content. Learn more at https://webaim.org/techniques/skipnav.
+//
+// To test this component's funtionality, place your cursor on "Example" and click.
```
yvonnetangsu:
```suggestion
// To test this component's functionality, place your cursor on "Example" and click.
```

josephgknox: Thanks!

sherakama:
https://github.com/SU-SWS/decanter/pull/303 more comments here.

sherakama:
Three green checkboxes. Merging.

core/scss/components/skiplinks/_skiplinks.scss
```diff
@@ -0,0 +1,31 @@
...
+// Then, press the tab key. The Skiplinks link will slide into view at the top
+// lefthand corner of the screen. Press the return key (or click on the link) to
+// skip to the targeted element.
+//
+// Markup: <header>
+//   <a href="#main-content" class="su-skiplinks">Skip to main content</a>
+// </header>
+//  <main>
+//   <h2 id="main-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
+// </main>
```
JBCSU: @sherakama Hardcoding the markup here means the styleguide doesn't use our twig templates. In the other branch I refactored the twig files specifically so that the styleguide could use the roll-up `skiplinks.twig` file, which evaluates to essentially this hardcoded markup, while coding the individual elements separately (`link.twig` & `target.twig`) files so site builders could place the elements appropriately in their sites' markup. So is there still a reason to hardcode the markup here? I think having the styleguide use `skiplinks.twig`, which in turn uses the twig files we're recommending site builders use, is preferable.

sherakama: I was trying to find a way to balance effort, documentation, and keeping the twig templates directly usable vs an example. 

* This was a fast (lazy) way to create the correct output in the styleguide
* Maintained a usable twig file. 

I'd prefer to avoid having twig files merely to generate styleguide output as we can do that in the markup and to keep the intent of the twig files as something that an author can/should/could use directly in their project.

JBCSU: The style guide is how I test the twig files. If they're not actually used in the style guide, how do I know they work? I don't want to have to wait until I build a site on top of decanter to find out there's a problem with the templates.

PR #304, opened 2019-03-01
Add linear gradient background mixin

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Adding a linear gradient mixin (written by author in below link) 
https://www.sitepoint.com/building-linear-gradient-mixin-sass/

# Needed By (Date)
- Whenever

# Urgency
- not urgent

# Steps to Test

1. pull this branch
1. Use this mixin on a component in the styleguide, e.g., 
`@include linear-gradient(42deg, #b58234 0%, #d2b545 50%, #d7c04d 50.01%, $color-white 100%);`
2. Run` grunt styleguide` and see if it renders the expected gradient

- Decanter
- Scott's event calendar

- @stocker


core/scss/utilities/mixins/gradient/_linear-gradient.scss
```diff
@@ -0,0 +1,30 @@
...
+
+@mixin linear-gradient($direction, $color-stops...) {
+  @if is-direction-or-angle($direction) == false {
+    $color-stops: ($direction, $color-stops);
+    $direction: 180deg;
+  }
+
+  background: nth(nth($color-stops, 1), 1);
+  background: -webkit-linear-gradient(to-legacy-direction($direction), $color-stops);
+  background: linear-gradient($direction, $color-stops);
```
josephgknox: I'm thinking (I am assuming you are, too) that we can ignore the related CC issue here, as the prefix is there as a fallback/level of protection. We have similar "errors" on other components, including Alerts and others that call both the `.png` and .svg` image files as `background-image`. One related question: is the prefix still required for modern browsers? Perhaps we want it there anyway to help support older versions of Safari, etc.?

yvonnetangsu: @josephgknox That's a great question... I believe the author wrote the mixin a while back. I'm looking at https://caniuse.com/#feat=css-gradients, it looks like only Chrome v4-25, Safari 4-6 and iOS Safari 5-6.1 (the one ones that actually matter) and they all add up to <0.3% global usage. I think I might be ok with removing the -webkit support since gradients are fairly minor enhancements. What do you think? That way I can actually remove some of those helper functions as well.

yvonnetangsu:
John mentioned that Kerri is using gradients in recent designs and Scott's event calendar uses them, so I thought I'd just add it to Decanter 😃

core/scss/utilities/mixins/gradient/_linear-gradient.scss
josephgknox: I thought that might be the case. I'm not sure. I don't think we have an established threshold for these things, but I would imagine that <0.3% is below the threshold. I'm good with dropping it. Maybe we get one more vote? @sherakama?

josephgknox:
@yvonnetangsu yep! Kerri and I have been incorporating gradients in a few of our recent designs for clients. This is going to be really handy.

core/scss/utilities/mixins/gradient/_linear-gradient.scss
yvonnetangsu: Could we get a vote, @sherakama 😺 ? Ok to drop support for legacy -webkit convention? If ok, I'll get rid of the to-legacy-direction function too.

sherakama: Isn't that what our autoprefixer is supposed to take care of? I thought we were avoiding vendor prefixes in our SASS.

yvonnetangsu: With linear gradient it's a little trickier. It needs not only a prefix but the direction convention and angle calculation is different (which needs a helper function).

sherakama: Gotcha thanks,

Ok, our compatibility is listed here: https://github.com/SU-SWS/decanter/wiki/Compatibility

I think the browsers you mentioned are old enough that we don't have to support them. In this case, I am good with dropping the extra webkit option.

josephgknox: Woot! The vote is in. Soon to be another PR merged.

yvonnetangsu: @josephgknox @sherakama -webkit support removed, but I did keep the convert angle helper function since it might come in handy for other things later.

yvonnetangsu:
Sneaked in a z-index tweak for the styleguide brand bar. Somehow skiplink (on focus) got buried under it again - possibly something got mixed up in the multiple skiplinks branches yesterday 😄

josephgknox:
Thanks @yvonnetangsu! @sherakama this good by you, too?

PR #306, opened 2019-03-05
120-skiplinks-b: Alternative render.

sherakama:
@JBCSU @yvonnetangsu @josephgknox 

How about something like this. With this approach, the twig template is usable directly in your work and the rendered output has the example we want. 

Thoughts?

### DIFF
--- a/core/scss/components/skiplinks/_skiplinks.scss
+++ b/core/scss/components/skiplinks/_skiplinks.scss
@@ -11,10 +11,10 @@
 // of the main content. Learn more at https://webaim.org/techniques/skipnav.
 //
 // Markup: <header>
-//   <a href="#skip-links" class="su-skiplinks">Skip to content</a>
+//   <a href="#skip-to-content" class="su-skiplinks">Skip to content</a>
 // </header>
 //  <main>
-//   <h2 id="skip-links" class="su-sr-only-text" tabindex="-1">Main Content</h2>
+//   <h2 id="skip-to-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
 // </main>
 //
 //
--- a/core/scss/components/skiplinks/_skiplinks.scss
+++ b/core/scss/components/skiplinks/_skiplinks.scss
@@ -10,7 +10,13 @@
 // to the main content as sighted, mouse users.
 // of the main content. Learn more at https://webaim.org/techniques/skipnav.
 //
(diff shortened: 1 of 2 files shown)


yvonnetangsu:
> @JBCSU @yvonnetangsu @josephgknox
> 
> How about something like this. With this approach, the twig template is usable directly in your work and the rendered output has the example we want.
> 
> Thoughts?

I do like this approach! Btw, I just noticed this now, should we make this singular "skiplink" so it's consistent with all the other components?

core/scss/components/skiplinks/_skiplinks.scss
```diff
@@ -10,7 +10,13 @@
 // to the main content as sighted, mouse users.
 // of the main content. Learn more at https://webaim.org/techniques/skipnav.
 //
-// Markup: ../../templates/components/skiplinks/skiplinks.twig
+// Markup: <header>
+//   <a href="#skip-to-content" class="su-skiplinks">Skip to content</a>
```
yvonnetangsu:
```suggestion
//   <a href="#main-content" class="su-skiplinks">Skip to content</a>
```

core/scss/components/skiplinks/_skiplinks.scss
```diff
@@ -10,7 +10,13 @@
...
+// </header>
+//  <main>
+//   <h2 id="skip-to-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
```
yvonnetangsu:
```suggestion
//   <h2 id="main-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
```

josephgknox:
@sherakama this approach also works well for me. I'll merge this into the other branch. We can continue discussion there if needed.

@yvonnetangsu it's a good question regarding the component naming convention. I'd argue that this accessibility-related component might be OK to leave as plural, as the chances of this not skipping more than one link (used primarily for skipping navigation) is rare. Also, it appears to most commonly be referred to as Skiplinks on the inter-webs. Willing to discuss this more, too. I might be crazy. :)

yvonnetangsu:
@josephgknox I'm comfortable going with the most commonly known term 😸 Thanks for checking!

JBCSU:
@sherakama @josephgknox I just committed to this branch an alternative that I think works quite well. Please let me know what you think.

PR #307, opened 2019-03-05
120 skiplinks b

JBCSU:


JBCSU:
> You definitely put in the effort for this one. I'm totally ok with splitting this up in to separate templates as it gives a good 'working example' of how it can be done. A couple of comments for refinement however:
> 
> * Twig template names. Can we remove the underscore as that is not a twig convention but only a SASS convention?

Oh, yeah, sure. Sorry, I thought I had seen twig files named that way.

> * Can we restore the anchor naming work that Yvonne and Joe had done previously?

I don't think so, because:

- the id is now used in 2 separate templates,
- it needs to have the same default in both, and
- in normal usage the rollup .twig file will not be used, so we can't set the shared default there.

I therefore picked an ID I thought would be unique but not bizarre: `su-` to match decanter naming conventions; `main-content`, because that's semantically descirptive; `1891` is the year Stanford was founded. We can pick a different default, but I think we have to actually hardcode a default because of the above-stated reasons.

> * The skip to content options that I have seen before, jump to an anchor on a div instead of into a heading. Not having actually tested this, I would imagine that a screen reader's behavior would differ if we jumped to a heading and I am not sure the repercussions of this. Something to ask Jiatyan.

It's a heading so the screen reader will actually read it. We had this discussion for homesite, and that's why homesite uses an `<h2>`.

josephgknox:
@yvonnetangsu you good with this, too? :)

PR #330, opened 2019-03-06
Update _alert--info.scss

sherakama:
# READY FOR REVIEW

# Summary
- Links in the alert text show up as red on a dark blue background. This makes the colors more readable.

# Needed By (Date)
- ?

# Urgency
- Low

# Steps to Test

1. Check out this branch. 
2. Compile the CSS
3. Add a link to the `info` variant
4. Review colors and hover.

# See Also

### DIFF
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -6496,7 +6496,7 @@ dfn {
 
 .su-alert--info {
   background-color: #00548f;
-  background: url("../img/alert-info.png");
+  background-image: url("../img/alert-info.png");
   background-image: url("../img/alert-info.svg");
   color: #ffffff; }
   .su-alert--info a {
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -6487,11 +6487,11 @@ dfn {
     .su-alert .su-alert__body .su-alert__text {
       margin-top: 0;
       margin-bottom: 0; }
-    .su-alert .su-alert__body a, .su-alert .su-alert__body a:link {
-      color: #ffffff;
-      text-decoration: underline; }
-    .su-alert .su-alert__body a:hover {
-      color: #eaab00; }
+  .su-alert a, .su-alert a:link {
+    color: #ffffff;
+    text-decoration: underline; }
+  .su-alert a:hover {
+    color: #eaab00; }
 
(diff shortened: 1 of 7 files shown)


core/css/decanter.css
```diff
@@ -6496,9 +6496,14 @@ dfn {
 
 .su-alert--info {
   background-color: #00548f;
-  background-image: url("../img/alert-info.png");
+  background: url("../img/alert-info.png");
```
josephgknox: Making this `background:` instead of `background-image:` overrides the `background-color:`. Therefore the blue color is now appearing. Also, the image is repeating if we don't tell it not to.

josephgknox:
@sherakama compiled the CSS and addressed the whitespace CC issues. I have also left a comment regarding the background call change.

core/css/decanter.css
yvonnetangsu: @josephgknox Good catch! I just pulled down the branch and was wondering why everything was white 😺

sherakama:
Whoops, my bad. I didn't mean to leave that background change in there. Thanks and good catch. Ready for second pass.

yvonnetangsu:
I just realize I totally forgot about the default one (with the off white background) in my tweak. Will push up a fix.

yvonnetangsu:
Sorry about that - ready for another look. I leave the default one and the yellow one as default link color and applied @sherakama's color scheme to the 3 darker ones (green, blue and red - the red one really could use the new scheme since it's red link on red). I also added a link in the alert message in the json file since links in alerts are pretty common so it might be nice to see it in the styleguide. Let me know what you think - feel free to change things.

josephgknox:
All of the combinations pass AA color contrast for large text, which is defined as being 18.66px or larger. We're only above 18px when we hit our largest breakpoint. That said, I think we need to update some of these.

josephgknox:
I've updated the necessary link/hover colors so that they pass color contrast for fonts smaller than 18.66px (which now covers us for all breakpoints). The states are now certainly more subtle, but in order to stick with Stanford-branded colors as our default colors for this component, I think this is the best route.

josephgknox:
If you two agree, this is ready to merge.

PR #332, opened 2019-03-08
Refactor card component using placeholder and match Figma design

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Refactor the card component to use new modular-spacing mixin and make sure it looks the same as the mocks in Figma. Buttons in cards weren't showing up but it should now.
- ~~Create card mixin to allow passing in options - orientation, drop shadow depth, border,  headline font family and vertical alignment of content (for horizontal variant)~~
- Use %placeholder and @extend to reduce repetitive code
- Set some max-width for components (cards, CTAs) in the kss scss files so devs/designers visiting the site can get a better idea of what the components look like on a real site.

# Needed By (Date)
- Whenever

# Urgency
- Not urgent

# Steps to Test

1. Pull this branch
2. Run `grunt styleguide`
3. Check the styleguide that the default renders for both vertical and horizontal cards match the mocks in Figma. Check that the Hero component also looks like (it uses the card component).

- Decanter
- Event calendar

- Talked to @JohnHolleman about default CTA link and button styles (in general, not necessarily the ones inside the cards). He said CTA links should be semibold (font-weight: 600, which is actually same as variable $font-bold in Decanter). On the other hand, font-weight for text in buttons should be normal (font-weight: 400). Both are used in cards as well so I made the tweak in the Button and Link component.
- Due to the inherent line-height of our font stack, I had to reduce some of the modular-spacing steps (compared to what Figma specified) to match the spacing of the rendered cards and the mocks visually. I talked to @JohnHolleman about this and he said he would rather we match the design visually than having the modular step numbers match exactly.
- I took away the margins for the Button component so it's consistent with other components (no spacing on the outside of the component).


yvonnetangsu:
2 cc complaints with the TODO but I believe @sherakama put those in so I'm leaving them there for now 😄

sherakama:
Thanks for all of the little fixes on top of the card refactor in this one. I removed the TODO comment as the modular text is now in place and the todo is done. I'll avoid adding those in the future as this is a prime example of a poorly written todo.

A couple questions on the approach if you don't mind. Why the refactor in to a extends and a mxin? A card component seems like a complex and structure specific item that doesn't really fit a mixin. You also moved the body of what a card is to in to an extends. Do you envision other developers needing this body of code for their work? How would one re-use this extends outside of Decanter?

yvonnetangsu:
Those are great questions, @sherakama . I'll try my best to answer them. 

1. I went for the mixin approach so it can take in parameters. While looking at the cards on Scott's event calendar site, I see how it would be a pretty common thing for different sites to have slightly different card designs (eg, with/without borders or shadows, headline fonts). I like having clean CSS output so less site-specific overriding is always a plus for me, and creating a mixin with parameters allow for that.

2. As for why using @extend and the %placeholder class - while writing the mixin, I realized how they can generate a lot of repetitive code in the output. I did a bit of research and found that using %placeholder and @extend with @mixin eliminates that repetitiveness. Basically, if there are two variants, by using this method the generated CSS output would be:
.variant-a, .variant-b {big block of common code}
.variant-b {unique code to variant b}

Instead of (repetitive!!)
.variant a {big block of common code}
.variant b {big block of common code + unique code to variant b}

Some of the readings I did:
http://krasimirtsonev.com/blog/article/SASS-mixins-extends-and-placeholders-differences-use-cases
https://blog.teamtreehouse.com/extending-placeholder-selectors-with-sass

So it really has nothing to do with reusing the code outside the Decanter, just want to keep our CSS dry 😄 

3. For the TODOs - actually they're still semi valid. We have @modular-typography but we don't have those header mixins, eg, @display-a etc. I know JB's lever/typography work is still ongoing so I didn't want to create those yet as it might interfere with what she was working on. So, at some point they'll still need to be tweaked but should be easy 😄 

Hope that answer your questions, and let me know if you have other concerns!

josephgknox:
+1 to the thanks for addition fixes on top of the central work on this PR.

My initial thought was that of Shea's: a card is a more complex, structure-specific component that doesn't benefit from being its own mixin in the same way as other elements/[atomic]components. That said, it appears as though you've done your research and are able to defend this choice in a way that makes sense to me.

If this is the route we decide to go, would we want to ensure we are consistent with this approach? For example, the button component could likely be updated similarly (using `@extends`, etc.).

yvonnetangsu:
@josephgknox, yeah, I was thinking that some other components could use this structure. I'd like to add a variant for main nav to be used on dark backgrounds (eg, on the styleguide site), and thought it would be easier if there's a main-nav mixin that allows me to pass color variables into fancy-hover instead of manually overriding the default colors. It would be nice to have a consistent approach but there's probably no rush - maybe go back and tweak when we have time and keep in mind for future components?

sherakama:
Being DRY is very good but there is a balance and trade in SASS. In this approach, you are trading DRY SCSS for duplicate code in the CSS. Every time you call this mixin you are duplicating the code in the extends and are adding unnecessary bloat to the rendered CSS. 

That aside, when you start going down the parameter route for creating slight differences in a component what you are creating is actually a variant. Our approach thus far is to be very explicit with the variants available and to not open up parameters that would create exponential complexity within the Decanter project. I would expect other projects to extend the default su-card class if they needed to create variants within their site. 

This sounds like a good topic for a zoom meeting.

yvonnetangsu:
@sherakama, yeah, it's true that it adds some redundancy in the form of adding the selectors for the variants even though we don't need them (e.g., we don't need `, .component-variant` below since we already include the `.component-default` class in all variants's markup):
```
.component-default, .component-variant {
  big block of common code
}
```
With that said, I wonder if I can propose the following:

- Write the mixin, but get rid of the parameters that are related to the variants, eg, for cards, the mixin would be @card($stroke, $shadow, $headline, $content-valign) and get rid of the $orientation parameter which renders the horizontal variant.
- This way, I'm not calling the mixin again in the variants, instead, I'm only calling it in the default option, and thus it won't generate extra code in the rendered CSS file.
- This is similar to how we've been creating the Button components. We have a button mixin, and call it in the default. The variants just have extra scss.
- The parameters allow for some small variations (different from the bigger variants) and the site builders would not have to dig into the code to see how to change things, e.g., the $content-valign allow someone to center aligned the horizontal card and the content quickly.

I know JB prefers having mixins available for components, and this way, it is still possible to have some flexibility with the parameters if one chooses to use the mixin instead of the component. The output CSS will basically stay the same. What do you think?

yvonnetangsu:
@shea, I just pushed up a tweak based on what I proposed above. Is this better? It doesn't generate the redundant markup in the selectors anymore. All the new CSS added are just due to the new code on card elements to match the Figma docs (mostly due to the modular-spacing on individual elements).

And here's another question, since some of the modular-spacing-base are the same for some breakpoints (`xs` and `sm` share one;  `md, lg, xl` share another one). Do you think we should write out the media queries with just those 3 ranges, instead of using a automatically for loop to pull from the map which generate 6 queries? 

The for loop might be easier to maintain, in case we change the base for some breakpoints in the future, but the consolidation would reduce our redundant CSS by a lot (eliminate 3 media queries for each element that we use @modular-spacing on). What do you think?

sherakama:
I'm on board for putting the base CSS styles for a component into an extends as that will help keep both the SCSS and CSS DRY. However, I still have concerns about the mixin in Decanter and how it will compete with variants. Let's talk about this together. I will through a zoom chat on your calendar.

yvonnetangsu:
> I'm on board for putting the base CSS styles for a component into an extends as that will help keep both the SCSS and CSS DRY. However, I still have concerns about the mixin in Decanter and how it will compete with variants. Let's talk about this together. I will throw a zoom chat on your calendar.

Yeah, I think having some common base styles in %placeholder class and using those with @extend would be a good start. Thanks for sending the invite!

yvonnetangsu:
@sherakama @josephgknox , card has been refactored using placeholder and extend - and no more card mixins. Let me know if this looks ok now 😃

yvonnetangsu:
@josephgknox brought up a good point - a card should be allowed to only link to one URL (eg, we shouldn't allow the button to link to one, and an action link to link to another URL). I pushed up a small tweak so there can only be one link href variable. 

I started a discussion on Slack about whether we should add some booleans to enable linking the headline or image (since both button/action link are optional fields in the card), or even linking the whole card. Please let me know what you all think (@josephgknox @sherakama @JBCSU ). However, maybe that's for another PR to avoid scope creep 😂

sherakama:
* One link is definitely preferred. How it is applied should be configurable.
* Wrapping the whole card in a link is definitely an option that I have seen in many sites
* Create a ticket for this as yes, this is scope creep.

josephgknox:
+1 to @sherakama comment on scope creep. Let's make another ticket for that. Thanks, again, for starting that discussion in Slack/here.

Issue #280, opened 2019-01-16
Implement webpack


sherakama:
Goals:
 - Compile all javascripts into one file
 - Ability to pick and choose which files the developer wants
 - Define and document file structure & organization
 - Implement linters
 - What version of EMCA Script? ES6.
 - (investigate) Replace grunt compilers for SCSS with webpack compiler (optional)

PR #334, opened 2019-03-12
90 basic webpack

JBCSU:
# READY FOR REVIEW
Have at it, folks.

# Summary
- replace the grunt build system with webpack

# Needed By (Date)
- 3/27/2019 (5.0 release)

# Urgency
- moderate - site builders' work will be easier when this is finished

# Noteworthy changes
- npm has been upgraded to the latest version supported by nvm, 10.15.1. Make sure you have this version installed via nvm: `nvm install 10.15.1` .
- Almost all npm packages have been removed. Many new ones have been added. All are now at their latest version.
- Generated css and js now lives in `core/dist/css` and `core/dist/js`, respectively. (They used to live in `core/css` and `core/js`.)
- Images referred to in `.scss` are now inlined in the generated`.css` as data.
- Images referred to in `.twig` files are (mostly) merely placeholders, and therefore, IMO, are not technically part of core. I have placed those images in a new root directory, `img-placeholders`. All images in that directory are copied to `styleguide/img` whenever you (re-)build the styleguide. Templates that refer to placeholders now refer to `/img/...` (absolute vs. relative url). (The one exception is the search icon in the search component. I propose the search icon be implemented via scss so it can be inlined / bundled with core.)

# Known Issues
- CSS source maps aren't generating. No idea why. I'll keep trying.
- linters are not yet configured. That should be relatively quick, and I can probably get it in this PR.
- I have not (yet) provided an equivalent to `browserSync`. `npm run watch` will rebuild the `.css` and / or `.js` when corresponding source files change; however, you have to manually refresh your browser to load them. I believe I can implement this with webpack's [hot module replacement](https://webpack.js.org/concepts/hot-module-replacement/), but that will take some time to configure properly. I would like to address that in separate PR.
- There is no watch on `.twig` or `.json` files, so you need to manually run `npm run styleguide` and refresh your browser when those files change. I don't think there's a way to implement this purely with npm (webpack is not involved in kss generation). Do we need to retain grunt for this?
- I have not (yet) provided an equivalent to `grunt deploy`. I haven't found an npm equivalent for [grunt-deploy-site](https://www.npmjs.com/package/grunt-deploy-site). Maybe we need to retain grunt for that one task???
- There are some codeclimate complaints. I believe most of these are issues we agreed to accept when we merged the nav branch, but we can discuss.

# Steps to Test

:warning:
**WARNING:** This PR changes (almost) every build tool we use, including the version of node / npm. It also changes where output files are stored (from `core` to `core/dist`). It is challenging to switch between this branch and any other branch in the repo. You may want to clone this branch to a separate local repo on your machine to avoid the overhead of switching back and forth.
:warning:

## One time only

- If you don't already have nvm's latest version of node / npm, do `nvm install 10.15.1`.

## If you're testing in a pre-existing local repo:

1. `rm -rf node_modules` to wipe out all tools used in the grunt-based build env
1. `npm cache clean --force` to really wipe out everything
1. `rm -rf styleguide` to delete the assets generated by webpack (just for completeness)

## Once you're In a clean repo (either a new one, or one you cleaned following the preceding steps):

1. `nvm use` to use the version of npm specified in `.nvmrc`, i.e. 10.15.1
1. `npm install` to install the webpack tools

## To build decanter and the styleguide:

1. Build the assets and the styleguide
    1. `npm run build` to generate the dev version of css and js, and build the styleguide **OR**
    1. `npm run dist` to generate the prod version of css and js, and build the styleguide
1. You can also just build just the styleguide without rebuilding assets:
    1. `npm run styleguide` to generate the styleguide **NB** this does *not* build the css

## If you decided to do this in the same local repo you use for other decanter development and you want to switch to a different branch

Before you switch to a different branch you should:

1. `rm -rf node_modules` to wipe out all tools used in this branch
1. `npm cache clean --force` to really wipe out everything
1. `rm -rf styleguide` to delete the assets generated by webpack (just for completeness)
1. check out the branch you want to switch to
1. `nvm use` to use the version of npm specified in `.nvmrc`, i.e. 9.10.1
1. `npm install` to re-install the grunt tools

- Final assets have moved from `core/css/` and `core/js/` to `core/dist/css` and `core/dist/js`. All products built on Decanter that load the pre-built assets will need to look in the new directories.

- Issue #280

# See Also


JBCSU:
Oh. Looks like we used grunt to build the styleguide's custom css. Oy. :upside_down_face:

webpack.config.js
```diff
@@ -0,0 +1,152 @@
+const path = require('path');
+
+const assetDir  = './core/';
```
sherakama: What about changing this to `src`?

webpack.config.js
```diff
@@ -0,0 +1,152 @@
...
+const outputDir = 'core/dist';
```
sherakama: How about we put this at the root of the directory so that this isn't nested?

sherakama: eg: `const outputDir = 'dist';`

img-placeholders/alert-info.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sherakama: What is this new image directory for? Why can we not use the images in `core/img`?

webpack.config.js
JBCSU: My thought was to have everything someone building on top of decanter might need contained in one directory. We had settled on `core` a while ago for consistency with other npm packages, specifically bourbon. If people want to cherry pick components, they can find source in `core/scss` and `core/js`; if they just want to use the prebuilt assets, they can find them in `core/dist`.

I suppose this breaks down a bit because people building on decanter might want to use the twig templates, which are not in `core/`. But my goal was for folks to be able to everything they need to build with in `core/`.

img-placeholders/alert-info.svg
JBCSU: If you're talking about the `stylesheet/image/` directory, that's for placeholder images that are required for the stylesheet to look good but are not required (or even recommended) for someone building on top of decanter.

For example, `hero-banner.png` is a necessary placeholder for the hero component in the styleguide, but no one building on top of decanter needs (or wants) that specific image.

Contrast that with `caret-down-black.svg`, which is a necessary part of the nav component, so people building on top of decanter must have that specific image as part of their build.

webpack.config.js
```diff
@@ -0,0 +1,152 @@
...
+    return recursiveIssuer(module.issuer);
+  } else if (module.name) {
+    return module.name;
+  } else {
+    return false;
+  }
+}
+
+module.exports = {
+  entry: {
```
sherakama:
```suggestion
  devtool: 'source-map',
  entry: {
```

Try this out to enable source maps along with enabling css sourcemaps in the CSS Loader.

webpack.config.js
```diff
@@ -0,0 +1,152 @@
...
+            presets: [ '@babel/preset-env' ]
+          }
+        }
+      },
+      {
+        test: /\.s[ac]ss$/,
+        use: [
+          {
+            loader: MiniCssExtractPlugin.loader,
+            options: {}
```
sherakama:
```suggestion
            options: { sourceMap: true }
```

img-placeholders/alert-info.svg
sherakama: I was referring to the `img-placeholders` directory. What is this one for?

`core/img` is for assets required by decanter and `kss/builder/decanter/kss-assests` should be for all assets that are just for the styleguide and not for the decanter package. KSS node moves all files in this directory in to the styleguide build. In the non-webpack build I had an extra copy step to move the Decanter core package in to styleguide instead of duplicating them in the kss-assets directory.

webpack.config.js
sherakama: I think I am leaning more towards the `src/dist` model these days. I've seen that pattern in a number of packages as well and seems to be the module most webpack projects use. 

https://stackoverflow.com/questions/23730882/what-is-the-role-of-src-and-dist-folders/23731040#23731040

https://guide.freecodecamp.org/miscellaneous/react-with-webpack-file-structure/

We can add a post-build step to move the twig files into place if we feel that is neccessary but those files, along with the sass, feel like source files to me. I think at some point we will get into a javascript framework where the templates would be outputted to this dist directory.

@yvonnetangsu @josephgknox thoughts?

webpack.config.js
```diff
@@ -0,0 +1,152 @@
...
+              url: true
+            }
+          },
+          {
+            loader: 'postcss-loader',
+            options: {
+              sourceMap: true,
+              plugins: () => [
+                autoprefixer( {
+                  browsers: [ 'last 2 versions', '> 1%' ]
```
sherakama: This matches the configuration we had with the grunt compiler.
```suggestion
                  browsers: [ 'last 2 versions', 'ie 11' ]
```

webpack.config.js
yvonnetangsu: Hmm.... I do like the idea of having separate src/dist directories because there's no ambiguity and that the asset folders (js, css/scss) have the same depths in the nested structure. In this model, where would the img folder go? Twig files seems like they belong to both. I also like the idea of someone developing on top of decanter can find everything they need in one directory. 

I see that @sherakama just pushed up an alt branch - I'll check it out.

img-placeholders/alert-info.svg
JBCSU: Shea moved the images to the right place.

webpack.config.js
JBCSU: Shea applied this change. Thanks!

JBCSU:
@sherakama I've committed all the changes we discussed today:

- reorganized the files into `core/src` and `core/dist`
- reorganized the webpack plugins, inspired by your `90-fixup`, which resolved the problem with the clean plugin
- got `devMode` working (triggered off the name of the npm script: 'dist' = production, anything else is dev)
- replaced the clean plugin with the file manager plugin, and used it to
  - clean generated assets before building
  - copy generated assets to the styleguide after building (there's a comment explaining why I did it the way I did)
- merged `master` to get the search icon inlined as data

Thanks again for your help on this PR in general, and today in particular. I look forward to discussing tomorrow.

Issue #333, opened 2019-03-08
Add main nav variant for used on dark backgrounds


yvonnetangsu:
Related to "Replace KSS main nav with main nav component"
https://github.com/SU-SWS/decanter/issues/328

We need a variant for main nav for it to look good on dark backgrounds, e.g., the styleguide. Possibly refactor main nav into a mixin so it can take colors as parameters.

yvonnetangsu:
https://github.com/SU-SWS/decanter/pull/332

yvonnetangsu:
Merged #344

Issue #328, opened 2019-03-06
Replace KSS main nav with the main navigation component.


sherakama:
Now that we have a main navigation component, replace the current navigation with the component.

yvonnetangsu:
I'm assigning this to myself so I can see how it behaves when used in real life and make adjustments as needed.

PR #344, opened 2019-03-15
Main Nav Light on Dark Background Color Variant

yvonnetangsu:
# READY FOR REVIEW

# Summary
- A variant of main nav to be used on dark backgrounds, e.g., the Decanter site
- The only difference between this variant and the default are the first level links on desktop, and the mobile hamburger button. The dropdown/mobile version are the same as the default styles.
- I was able to show it to @JohnHolleman before he left for long weekend. It has his blessing 😄 

# Needed By (Date)
- Maybe before v5 release

# Urgency
- N/A

# Steps to Test

2. Look at the .su-main-nav--dark variant and see if the look and behavior are ok

- Decanter

- https://github.com/SU-SWS/decanter/issues/328
- https://github.com/SU-SWS/decanter/issues/333


core/scss/components/main-nav/_main-nav.scss
```diff
@@ -30,7 +33,7 @@
 // Style guide: Components.Main Navigation
 //
 
-$custom-active: #ec0513; //custom high contrast active/current link border for accessibility
+$high-contrast-red: #ec0513; //custom high contrast active/current link border for accessibility
```
sherakama: if we are going to add variables to components we should prefix them with the component name so we don't run into global naming collisions.

sherakama: Otherwise, roll this back up to the colors variable file.

yvonnetangsu:
> This looks great. One comment on the variable name for the new red and one accessibility check. I think SVGs are supposed to have a <title> tag in them.

I added that new red into the global color palette after talking to @JohnHolleman. It's now called $color-vivid-red and it's in our swatches section in Core. 

As for the SVG, it seems like a topic that needs some discussion, eg, a caret in one component might mean one thing and mean something else in another. For the main nav they're used as background-images, and we have aria-expanded to indicate the states which mean the same thing as the carets and plus/minus signs. On the hambuger buttons, we have actual text in the button (menu/close) so maybe we're fine for now? What do you think?

yvonnetangsu:
@josephgknox - What do you think? We could use your design-developer eyes 😄

josephgknox:
@yvonnetangsu this looks great! More variants FTW! One questions/comment: should we add left/right padding to the dark variant? Feels tight.

yvonnetangsu:
> @yvonnetangsu this looks great! More variants FTW! One questions/comment: should we add left/right padding to the dark variant? Feels tight.

Hey @josephgknox - do you mean left/right padding so there's more spacing between each link (1st level desktop) or more spacing to the whole menu bar (also 1st level desktop)? If it's the latter, no worries, it doesn't actually have a background color. The dark background is just for demo and on the styleguide div only 😃  But maybe we should add some padding for the demo so it looks better anyway?

josephgknox:
@yvonnetangsu I meant the latter. So, if this is primarily (only) for the hamburger, then the padding is a moot. Thanks for clarifying. I don't think this needs anything beyond what you have for demo purposes.

yvonnetangsu:
Thanks for the review and merging! I'll see if I can get the new nav working for the Decanter site.

PR #345, opened 2019-03-15
Reorganize files for CTA so it's consistent with other components

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Other components have variant styles in their own separate scss files. This PR makes CTA consistent with other components by using the same file structure.
- Some small cleanup. This component was my first PR - it could use some improvement.
- Create a centering mixin (recommended by many sites as a good utility mixin) and use it in this component (the `.cta--button-center` variant).

# Needed By (Date)
- Hopefully before v.5 release so everything's consistent

# Urgency
- Not urgent

# Steps to Test
2. Check that all the CTA variants are still the same.
3. Check that the @centering mixin is working.

- Decanter


sherakama:
This looks good. Thanks for this work. Perhaps move the mixin to the layouts directory instead of creating a new folder. May as well start to fill out those directories.

yvonnetangsu:
> This looks good. Thanks for this work. Perhaps move the mixin to the layouts directory instead of creating a new folder. May as well start to fill out those directories.

@sherakama, @centering has been moved under Layout. Let me know if you approve 😄

PR #347, opened 2019-03-18
Use %placeholder and @extend on button mixins to see if it makes code less repetitive

yvonnetangsu:
# READY FOR REVIEW

# Summary
- We will have a discussion on using @extend on some base styles when using mixins. This is a test to see if this is a good approach and what can be done differently.
- I picked the button mixin to test because it's simple and get used by other components a lot, and it has a few variants.
- Created a `%button-base` class with the button base styles that are mostly common to all button variants, and also used by some more complex components such as the CTA (uses the `.su-button--big` variant).
- @extend this `%button-base` class in the variants' mixins and see if it reduces our output CSS file size and scss file size.

# Needed By (Date)
- Maybe take a look before Zoom meeting on Wed?

# Urgency
- N/A

# Steps to Test
2. Check in the components Button, Link, CTA section that all the buttons still look and behave the same as on the live site.
3. Look at the files changed tab on this GitHub page and see what's deleted and what's added in the output CSS and scss styles.

- Decanter

- If this get merged we'll have to make sure the card and hero components stay the same or we'll have to do some small refactor.


yvonnetangsu:
Notes:

If we don't use the button components in the more complex components, e.g., CTA and cards, we can get away with not doing `@extend %button-base;` in the button variant mixins (@button-big and @button-secondary). However, we do use @button-big as standalone for e.g., the CTA button, and the CTA button doesn't have a .su-button class which button base styles - which is why I still include the `@extend %button-base;` in the button variant mixins.

core/scss/utilities/mixins/button/_button-primary.scss
```diff
@@ -7,17 +7,14 @@
 //
 // Style guide: Mixins.Button.button-primary
 //
-@mixin button-primary {
-  @include font-smoothing;
-  @include margin(0.5em 0.5em 0.5em null);
-  @include padding(1rem 2rem 1.15rem);
 
+%button-base {
```
sherakama: I like where you are going for this. I vote for moving this to its own file and documenting it like we do mixins. Perhaps we need to create a `/core/scss/utilties/placeholders/` directory?

yvonnetangsu: Yeah, I think that makes sense - do you think it's better to add a directory for each component that has placeholders or just all in the root /placeholders?

sherakama: Directories are good. We will likely get to a point where we will want them so let's start with them.

yvonnetangsu: @sherakama , I just pushed up a commit. I created `/placeholder` inside `/utilities`. I'm using the same structure as `/variables` by creating a `/components` inside `/placeholders` and moved the button placeholder style there. What do you think? Since we're most likely to have only 1 placeholder for each component, I didn't create a separate `/button` inside `/components`.

TL;DR - we now have
/utilities/placeholders
/utilities/placeholders/components/_button.scss (where new %button-base is)

What do you think?

yvonnetangsu:
> I made one more big change to this. I moved the CTA mixin to a placeholder as I feel that makes more sense than a mixin. I would say we would want to avoid complex components as mixins.

Yeah, I think this change makes sense. I noticed that the CTA buttons aren't working right now but I'm assuming you're working on it (says fixup wip in your commit message) so I won't touch it for now. Thanks for the tweak!

josephgknox:
Thank you two for you discussion and work on this ticket. Structure and strategy sound and look solid. One thing that we might want to address before merging this is:

For the CTA component: The button style is not taking. It reads "Button Text" but it is not a button. Also, for the `.su-cta--button-center` variant, the button is not appearing at all.

josephgknox:
The button not working on CTAs appears to be an order of operations issue:

```
Fatal error: ".su-cta" failed to @extend "%cta-base".
The selector "%cta-base" was not found.
Use "@extend %cta-base !optional" if the extend should be able to fail.
```
`'placeholders/index'` would need to come before `'mixins/index'` but can't because we are using mixins in placeholders.

yvonnetangsu:
@josephgknox ...interesting....you found a paradox!! 😂

josephgknox:
@yvonnetangsu Haha! Yep, the `_cta.scss` file has a typo (maybe intentional) of `@extends` rather than the correct `@extend` which is why the error was not originally being thrown. Not sure which route to go here. It would make sense to allow for mixins in placeholders, but if there is a back and forth of their needing each other, I'm not sure how to proceed without refactoring...

josephgknox:
@yvonnetangsu I also see you comment to Shea regarding the same/similar thing. Do we want to proceed with merging in this work, knowing that we will need to clean up the CTA buttons later?

yvonnetangsu:
> @yvonnetangsu Haha! Yep, the `_cta.scss` file has a typo (maybe intentional) of `@extends` rather than the correct `@extend` which is why the error was not originally being thrown. Not sure which route to go here. It would make sense to allow for mixins to be used in placeholders, but if there is a back and forth of their needing each other, I'm not sure how to proceed without refactoring...

Hehe..maybe this makes a case for keeping my original approach of @cta as mixin 😄 ! Just joking - I don't think @sherakama  likes this approach 🤔 

@josephgknox , That's ok, I'm totally fine with this not merged in before we know that everything's working. It was sort of an experiment for discussion on Zoom meeting tomorrow anyway.

josephgknox:
@yvonnetangsu sounds good! More tomorrow. :)

sherakama:
Sorry about the rollup mix up. I fixed the extends typo and included the cta appropriately. 

Pros for this approach:
* Placeholders reduce markup vs mixins
* Allows for developers to add their own wrapper classes and extend the placeholder

Cons for this approach:
* Adds fractioning and complexity to the code. 

Let's talk more about this tomorrow.

PR #350, opened 2019-03-21
Refactor site-search so the icon can be turned into inline CSS by Webpack

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Refactor the site-search component so that the `` tag that's wrapped by the submit `<button>` is replaced with screen reader text, and render the icon as a background image on the `<button>`. That will allow Webpack to parse the icon and embed it directly in the css.
- Add a max-width in the styleguide scss which makes it look more "realistic" for the styleguide site.

# Needed By (Date)
- Before v.5

# Urgency
- ASAP

# Steps to Test

2. Check that in the site search component, the magnifying glass icon, is functioning as a button and look the same as the one on the current website build.
3. Check if it meets accessibility requirements.

- Decanter

- #334 
- This is requested by @JBCSU due to her work on Webpack #334

### DIFF
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -8709,9 +8709,9 @@ a {
   -moz-osx-font-smoothing: grayscale;
   padding: 0.6rem 2rem;
   display: inline-block;
-  height: 40px;
+  height: 4rem;
   max-width: 100%;
-  border-radius: 30px;
+  border-radius: 3rem;
   font-size: 1.6rem; }
   .su-site-search__input::-webkit-input-placeholder {
     color: #2e2d29;
@@ -8743,10 +8743,10 @@ a {
   background: url("../img/icon-search.svg") no-repeat 0 0;
   opacity: .6;
   position: absolute;
-  top: 11px;
-  right: 12px;
-  width: 2.8rem;
-  height: 2.8rem; }
+  top: 1.1rem;
+  right: 1.2rem;
+  width: 24px;
+  height: 25px; }
   .su-site-search__submit:hover, .su-site-search__submit:active, .su-site-search__submit:focus {
     background-color: transparent;
     opacity: 1; }
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -8744,16 +8744,18 @@ a {
   margin-top: 0;
   margin-right: 0;
   margin-bottom: 0;
-  background-color: transparent;
+  background: url("../img/icon-search.svg") no-repeat 0 0;
   opacity: .6;
   position: absolute;
+  overflow: hidden;
+  width: auto;
+  height: 2.8rem;
   top: 11px;
-  right: 12px;
-  width: auto; }
-  .su-site-search__submit img {
-    display: inline-block;
-    max-width: 24px;
-    max-height: 24px; }
+  right: 12px; }
+  .su-site-search__submit p {
+    opacity: 0;
+    width: 3rem;
+    height: 2.8rem; }
   .su-site-search__submit:hover, .su-site-search__submit:active, .su-site-search__submit:focus {
     background-color: transparent;
     opacity: 1; }
--- a/core/css/decanter.css
+++ b/core/css/decanter.css
@@ -8743,15 +8743,10 @@ a {
   background: url("../img/icon-search.svg") no-repeat 0 0;
   opacity: .6;
   position: absolute;
-  overflow: hidden;
-  width: auto;
-  height: 2.8rem;
   top: 11px;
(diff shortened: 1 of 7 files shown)


core/scss/components/site-search/_site-search.scss
```diff
@@ -37,18 +37,13 @@
...
   right: 12px;
-  width: auto;
-
-  img {
-    display: inline-block;
-    max-width: 24px;
-    max-height: 24px;
-  }
+  width: 2.8rem;
+  height: 2.8rem;
```
JBCSU: The units for top & right vs width & height seem switched to me. The units for width and height are relative (to the document's font-size), but the image has a fixed width (24) and height).

Conversely, the units for top and right are fixed (in px), but the size of the input box (at least the width) might conceivably change. It appears the height is fixed at 40px, which I now question: if low-vision users set their base font size very large, will they be able to see what they type in the search field?

Anyway, are we sure we're using the right units for top / right vs. width / height?

core/templates/components/site-search/site-search.twig
```diff
@@ -24,7 +23,7 @@
  <form action="{{ action }}" method="{{ method }}" accept-charset="UTF-8">
   <label class="su-site-search__sr-label" for="{{ search_input_name|default('search-field') }}">{{ search_label|default('Search this site') }}</label>
   <input {{- input_attributes }} type="text" id="{{ search_input_name|default('search-field') }}" name="{{ search_input_name|default('search-field') }}" class="su-site-search__input" placeholder="{{- placeholder -}}" maxlength="128">
-  <button {{- button_attributes }} type="submit" name="submit" class="su-site-search__submit" aria-label="Search"><img src="{{ search_icon_src }}" alt="Search button" /></button>
+  <button {{- button_attributes }} type="submit" name="submit" class="su-site-search__submit su-sr-only-text" aria-label="Search">Submit Search</button>
```
JBCSU: Interesting application of `su-sr-only-text`: It seems to move the text within the `<button>` but not the button itself. That's super cool if it works cross-browser. Does it work in all supported browsers?

kss/builder/decanter/scss/_site-search.scss
```diff
@@ -0,0 +1,9 @@
+@charset 'UTF-8';
+
+// The site-search component
+
+#kssref-components-site-search {
+  .kss-modifier__example {
+    max-width: 40rem;
```
JBCSU: That does look better in the styleguide. I wonder, tho, if it might mislead people into thinking that the site search has a max width by default, and therefore they don't have to specify a width for it? If we changed `.kss-modifier__example` to `.kss-modifier__wrapper`, then the example box would be smaller, and it might be more clear to users that the site search component does not specify a (max-)width.

yvonnetangsu: That's a good point. I can certainly do that.

core/templates/components/site-search/site-search.twig
yvonnetangsu: Yup I've checked as far back as IE 10.

JBCSU: Awesome! Neat trick!

core/scss/components/site-search/_site-search.scss
yvonnetangsu: Let me check - it could be that svg was finicky. In some cases they appear ok with px values but jagged with em/rem values. It's also an earlier PR so it's possible that I could improve some of these things.

core/css/decanter.css
```diff
@@ -8740,16 +8740,13 @@ a {
   margin-top: 0;
   margin-right: 0;
   margin-bottom: 0;
-  background-color: transparent;
+  background: url("../img/icon-search.svg") no-repeat 0 0;
```
JBCSU:
```suggestion
  background: url("#{$image-path}/icon-search.svg") no-repeat 0 0;
```
I think this should refer to the $image-path variable.

sherakama: This is the rendered CSS. The SCSS does have the correct variable.

josephgknox: +1 Was just typing that. It looks like the SCSS is using the variable correctly.

yvonnetangsu:
> I added variables for the button name and text values as well as prepended the input_attributes and button_attributes with `search_` for naming consistency.
> 
> This looks good to me.

Thanks for the changes - look good!

core/css/decanter.css
JBCSU: Oops, my bad. Sorry 'bout that.

PR #361, opened 2019-03-27
90 cleanup

sherakama:
# READY FOR REVIEW
* Normalizes paths with `path.resolve`
* Adds webpack cli tool
* Adds webpack dev server
* Adds to file manager to copy templates into dist dir.


webpack.config.js
```diff
@@ -17,13 +16,14 @@ const ExtraWatchWebpackPlugin = require("extra-watch-webpack-plugin");
 
 // Paths
 const npmPackage = 'node_modules/';
-const srcDir = process.env.npm_package_srcDir;
-const outputDir = process.env.npm_package_distDir;
-const kssSrcDir = process.env.npm_package_kssSrcDir;
+const srcDir = path.resolve( __dirname, process.env.npm_package_srcDir );
```
JBCSU: i think it's safest to leave the srcDir as a relative path. If it's absolute, I think there's a risk of absolute paths ending up in our generated assets, e.g. references to background images that are too big to be inlined as data.

webpack.config.js
```diff
@@ -158,12 +164,6 @@ let coreConfig = Object.assign({}, config, {
...
           enforce: true
-        },
-        'kss': {
-          name: 'kss',
-          test: ( module, chunks, entry = 'kss' ) => module.constructor.name === 'CssModule' && recursiveIssuer( module ) === entry,
-          chunks: 'all',
-          enforce: true
         }
       }
     }
```
JBCSU: Good catch! Thanks!

package.json
```diff
@@ -5,16 +5,17 @@
...
+  "distDir": "core/dist/",
+  "kssSrcDir": "kss/builder/decanter/",
+  "kssDistDir": "styleguide/",
   "scripts": {
     "build": "webpack --mode=development",
     "dist": "webpack --mode=production",
     "styleguide": "kss --config=kss/kss-config.json",
     "watch": "webpack --mode=development --watch",
-    "dev": "npm run build && npm run styleguide && npm run watch"
+    "start": "webpack-dev-server --open",
```
JBCSU:
```suggestion
    "start": "webpack-dev-server --mode=development --open",
```
Adding the mode will suppress the warning about mode not being set.

webpack.config.js
JBCSU: Nevermind.

Issue #201, opened 2018-08-15
Create Date Stacked component


josephgknox:

josephgknox:
<img width="215" alt="atom_date-stacked" src="https://user-images.githubusercontent.com/8933670/44177645-ab5e5680-a0a3-11e8-9596-e737a3086835.png">

sherakama:
Scott wants it, Joe and Mike have it.

sherakama:
Needs refactoring. Please connect to work on it.

buttonwillowsix:
Yvonne confirms this work is in master.

PR #373, opened 2019-03-29
date stacked component initial commit

josephgknox:
# READY FOR REVIEW

# Summary
- Adding date stacked component

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Compile styleguide
3. Review design of date stacked component and related code

- Decanter

- #201 

# See Also


core/src/scss/components/date-stacked/_date-stacked.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  .su-date-month {
+    @include modular-typography(1);
+    font-weight: $font-x-light;
+    line-height: $displays-line-height;
+    text-transform: uppercase;
+  }
+
+  .su-date-day {
+    font-weight: $font-regular;
```
cjwest: How about using $font-bold?

core/src/templates/components/date-stacked/date-stacked.twig
```diff
@@ -0,0 +1,24 @@
...
+ * @file
+ * Date Stacked Component.
+ *
+ * Abbreviated month stacked on top of numerical day of the month.
+ *
+ * Available variables:
+ * - attributes: For additional HTML attributes not already provided.
+ * - modifier_class: Additional css classes to change look and behaviour.
+ * - month_of_year: The abbreviated month of the year.
+ * - day_of_month: The day of the month in digit form.
```
cjwest:
```suggestion
 * - day_of_month: The day of the month in 2 digit form.
```

core/src/scss/components/date-stacked/_date-stacked.scss
josephgknox: I like regular better.

core/src/templates/components/date-stacked/date-stacked.twig
josephgknox: 1 digit is what we want actually.

josephgknox:
@kerri-augenstein has provided designs, which can be seen at https://www.figma.com/file/Kmd4utmJFPRMVeCFEEBQhLtx/Stanford-Design-Library?node-id=2264%3A3774

core/src/scss/components/date-stacked/_date-stacked.scss
```diff
@@ -0,0 +1,28 @@
...
+//
+// Style guide: Components.Date Stacked
+//
+.su-date-stacked {
+  @include modular-spacing('padding', 0);
+  background: $color-black;
+  color: $color-white;
+  text-align: center;
+
+  .su-date-month {
```
JBCSU: I think the BEM naming conventions suggest that this should be named `.su-date-stacked__month`. For brevity, I might be ok with `.su-date__month`. Depends on if we're ever going to have a date that isn't stacked.

core/src/scss/components/date-stacked/_date-stacked.scss
```diff
@@ -0,0 +1,28 @@
...
+    @include big-paragraph;
+  }
+
+  .su-date-day {
```
JBCSU: Again, BEM naming conventions suggest that this should be named `.su-date-stacked__day`.

kss/builder/decanter/scss/_date-stacked.scss
```diff
@@ -0,0 +1,15 @@
...
+
+    @include grid-media-only('md') {
+      max-width: 101px;
+    }
+
+    @include grid-media-only('2xl') {
+      max-width: 105px;
+    }
+  }
+}
```
JBCSU: Should these max-widths be limited to the styleguide's rendering? Don't we want a max-width on the actual rendering?

core/src/templates/components/date-stacked/date-stacked.twig
```diff
@@ -0,0 +1,24 @@
...
+<div{{ attributes }} class="su-date-stacked {{ modifier_class }}">
+  {% if month_of_year is not empty %}
+    <div class="su-date-month">{{ month_of_year }}</div>
+  {% endif %}
+
+  {% if day_of_month is not empty %}
+    <div class="su-date-day">{{ day_of_month }}</div>
+  {% endif %}
+
+</div>
```
JBCSU: Do we want to change the container from a `<div>` to a `<time datetime="...">`? (See `<time>` on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time).) In that case we might also want a year. The year wouldn't (necessarily) render in the browser, but bots could recognize the actual date.

core/src/scss/components/date-stacked/_date-stacked.scss
josephgknox: You're right! This is an oversight by me. Thanks for catching this.

core/src/scss/components/date-stacked/_date-stacked.scss
josephgknox: Thanks!

kss/builder/decanter/scss/_date-stacked.scss
josephgknox: I thought a little about this one, and wasn't sure if the max-width should be tied to the component itself or just the style guide for display purposes only. I wondered if the components use case might dictate what width someone wants to set for it. That said, we might actually want a sensible default established; one that they can then override. I'll move this.

core/src/templates/components/date-stacked/date-stacked.twig
josephgknox: I'll explore this more.

sherakama: That looks handy. It appears it doesn't have IE support and we should still support IE 11 so my question is what would it look like in IE11? My guess is that it isn't supported but probably looks 100% fine and should be usable. I have been tricked before however.

kss/builder/decanter/scss/_date-stacked.scss
JBCSU: I do think the widths are inherent in the component. But maybe we could do this styling with a placeholder so people can override if necessary?

core/src/templates/components/date-stacked/date-stacked.twig
JBCSU: [caniuse](https://caniuse.com/#feat=html5semantic) > **Known issues** says:

> While the `time` and `data` elements can be used and work fine in all browsers, currently only Firefox and Edge 14+ recognize them officially as `HTMLTimeElement` and `HTMLDataElement` objects.

In fact, we use it on news.stanford.edu. Go to any story on that site and inspect the date the story was published - it's a `<time>` tag. We should of course test it in IE, but it should be fine.

I'm not at all clear on how much benefit it offers, but I'd still lean towards using `<time datetime="...">`.

kss/builder/decanter/scss/_date-stacked.scss
sherakama: That is a really good question. My initial knee jerk reaction is to leave components at 100% width with the idea that they will be used either in some sort of container or by a developer in which they could then set the widths. What are other style guides doing? Are they using atomic design principles? Are they applying widths to every level (atom -> organism)? I am sure somebody has gone through some learning lessons on this already and it would be good of us to find out.

sherakama: A couple of examples to digest: 
* http://walmartlabs.github.io/web-style-guide/#responsive-tile
* http://walmartlabs.github.io/web-style-guide/#grid-tile
* https://rizzo.lonelyplanet.com/styleguide/ui-components/cards#styleguide-block__item--1
* https://patterns.wvu.edu/patterns
*

core/src/scss/components/date-stacked/_date-stacked.scss
```diff
@@ -15,13 +15,22 @@
   @include modular-spacing('padding', 0);
   background: $color-black;
   color: $color-white;
+  max-width: 87px;
   text-align: center;
 
-  .su-date-month {
+  @include grid-media-only('md') {
+    max-width: 101px;
```
sherakama: Before you commit to this convention please do some research on the right approach on bounding widths in components. I'd like to know what others are doing and why.

josephgknox: Walmart: they are hardcoding width to many components, including cards, tiles, and buttons.

Bootstrap: Unless something is assigned a width of 100%, it's more rare that the framework hardcodes component widths. In the cases where it does (e.g., popover and tooltips) it has a corresponding variable (e.g., `$popover-max-width` and `$tooltip-arrow-width`).

Lonely Planet: they are hardcoding widths, and are even naming some components in direct relation to their width (e.g., "Double width card with cover image").

WVU: most components are inherently 100% width. They do hard code width of cards when inside of a component-like container.

U.S. Web Design System: most components are inherently 100% width. In the cases where it does assign a width, it has a corresponding variable.

IBM Carbon: assigns width to small (atom-level components).

josephgknox: There is certainly a mix of what others are doing. I'm in the camp of the middle-of-the-road approach, which I see as being: we can assign widths to smaller (atom-level components like the date stacked component), but pass the width through with a variable that can be more easily changed/overridden.

sherakama: Great, that sounds sensible to me. I have created you a ticket to document this decision and linked it to this thread for a later time.

core/src/templates/components/date-stacked/date-stacked.twig
JBCSU: I don't think it's meaningful to just have `{{ month_of_year}}` or `{{ day_of_motnh }}` in the `datetime` attribute - see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#Valid_datetime_Values). You could end up with `datetime="9"`. What would that mean?

I think we should do our best to  provide the most specific datetime possible, even if that mean requiring data that doesn't render visually. I think we should at least have `datetime="{{ year }}-{{ month_of_year}}-{{ day_of_month }}"`, which requires a year, even though no year renders visually. (Of course we'd need logic to build the string based on what data we actually have - we don't want dangling dashes.) We might also add some logic to include the time, if provided.

sherakama: I am for having the time tag in there with a useful datetime but I am unsure how to make anything required that doesn't visually impact the component. Perhaps we split the difference and provide the input variables for the proper datetime with month-day-year time and timezone and use Twig's date functions to default to the server environment: https://twig.symfony.com/doc/2.x/filters/date.html

yvonnetangsu: @sherakama  I'm looking at this:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

Seems like a yearless date string in the form of 12-30 for example is valid. Could we do something like <time datetime="{{ month }}-{{ day }}><span>{{ day }}</span><span>{{ month }}</span></time>?

core/src/scss/components/date-stacked/_date-stacked--no-background.scss
```diff
@@ -0,0 +1,11 @@
+@charset "UTF-8";
+
+//
+// Date Stacked No Background
+//
+
+.su-date-stacked--no-background {
+  @include padding(0);
+  background: none;
+  color: $color-black;
```
yvonnetangsu: If we're going to merge master into this branch, we could get the colors from the color map. We could leave it as variables for now also.

core/src/scss/components/date-stacked/_date-stacked.scss
yvonnetangsu: I'm fine with this max-width treatment.

core/src/templates/components/date-stacked/date-stacked.twig
sherakama: How about now?

yvonnetangsu: Ideally it would be nice to have just one `<time>` tag. Also, `datetime` with just the month alone might not be valid without a year attached? (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time). I'm fine with the current version though, except maybe only emit one `datetime` attribute so it doesn't repeat, if we're using 2 `<time>` tags? Other than that GTG.

sherakama: 🤔 hmmm. Perhaps we go back to the `<span>` tag for safety. I know it would be more semantic/accessible if the field used `<time>` but if we are to allow for a mix and match approach, I wouldn't want to have tag switching on the component as that could leave to confusing UI or behaviors if they alternated in a list or something.

yvonnetangsu: That's true....if you're anticipating cases that either the day or month might be missing, then `<span>` is a safer choice, since a valid time tag would require both a month and a day 🤔

core/src/scss/components/date-stacked/_date-stacked--no-background.scss
yvonnetangsu: Looks great - thanks for the change!

core/src/templates/components/date-stacked/date-stacked.twig
sherakama: Ok, back to plain ol' span until we come up with a better solution.

sherakama:
@yvonnetangsu Have a look and punch the green button when you are happy with the state of this PR.

yvonnetangsu:
> @yvonnetangsu Have a look and punch the green button when you are happy with the state of this PR.

GTG - merging!

PR #378, opened 2019-04-03
Webpack Revisions: Integration Improvements

sherakama:
# READY FOR REVIEW

# Summary
- I ran into troubles getting this to build with my Drupal theme so I needed a few changes. 
- Separate the JS, SASS, and KSS builds completely. 
- Change entry points to the SCSS files instead of having empty or dummy JS files. 
 - 

When trying to import and build using version 5 I ran into an issue with the JS file and the CSS files. I was not able to include and build them into my project independently. It may be possible to do what I want with splitchunks but I would rather keep the two things separate. 

# Needed By (Date)
- The sooner you can look at this the sooner it will help me.

# Urgency
- Not critical. 

# Steps to Test

1. Check out this branch
2. Throw out your `node_modules` folder
3. Run `npm install`
4. Run `npm run build` and ensure everything builds without error
5. Run `npm run styleguide` and `npm run start` and ensure the styleguide looks right
6. Read through the code and leave comments on the changes.

# See Also


sherakama:
@josephgknox @pookmish FYI & https://github.com/SU-SWS/stanford_basic/pull/81

webpack.config.js
```diff
@@ -249,30 +383,12 @@ let kssConfig = Object.assign({}, config, {
...
+  }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+// MODULE EXPORTS.
+////////////////////////////////////////////////////////////////////////////////
 
-// Module Exports.
-module.exports = [ coreConfig, kssConfig ];
+module.exports = [ DecanterJS, DecanterSass, DecanterKSS ];
```
JBCSU: I think that separating the js build from the sass build precludes my reason for introducing webpack in the first place: having completely self-contained components. In this configuration, I don't think someone building on top of decanter will be able to `@import` just the `main-nav` component. That component requires both .[s]css and .js, so, to get both in the same build, one of those files has to `@import` the other. But, because the js processing is now completely separate from the scss processing, neither one can `@import` the other and have a successful build. Am I correct in this?

What problem are you trying to solve by keeping the js and the scss separate? Should we talk?

sherakama: Perhaps I am approaching this wrong. 

What I would like to do is the following:
1. Add and render all of Decanter's javascript into one file along with my project's code.
2. Add and render all of Decanter's css into one file with a few variable overrides from my project  
3. Use decanter-no-markup in several other entry points with my project's variable overrides

https://github.com/SU-SWS/stanford_basic/pull/81/files?file-filters%5B%5D=.js&hide-deleted-files=true#diff-11e9f7f953edc64ba14b0cc350ae7b9dR35

If I am correct on my thinking, I would have to do some fancy split chunking to separate out the css from the JS when rendering my all decanter css file. 

Yes, this deserves some conferencing time and I'll put something on the calendar so we can work through it together. 

Thanks for the feedback.

JBCSU: Decanter's webpack config only affects generated files that end up in `dist/`. Those files are already processed, and therefore not appropriate to bundle into another project's assets. They're there for people who just want to have
```
<head>
...
<link rel='stylesheet' href='https://mysite.stanford.edu/assets/css/decanter.css' type='text/css' media='all' />
<script type='text/javascript' src='https://mysite.stanford.edu/assets/js/decanter.js'></script>
...
</head>
```
in their site's markup.

What gets built in your project is entirely independent of Decanter's webpack configuration - it depends solely on your project's webpack config. To accomplish what you want, you would `@import` the rollup files from decanter's `core/src`, not `core/dist`. And decanter's webpack config doesn't change anything in `core/src`.

If I'm still not understanding the problem, yeah, let's definitely talk. But I have Scott's events calendar project and our WordPress theme working with decanter#5.0.0, so I'm pretty sure we can do what you want with the existing config.

What we should change in the config, now that we have the filemanager plugin, is to delete files in `dist` that don't make any sense, e.g. `decanter-nomarkup.css`, which, by definition, is empty. I can open a separate PR for that.

sherakama: I am not relying on Decanter's webpack build for my project and I am using my own configuration in my project. The big change for me is this line here: https://github.com/SU-SWS/decanter/pull/378/files?file-filters%5B%5D=.js&hide-deleted-files=true#diff-141d53bc6cb5d14d1ccc375cf7ebdeb7L3 where I removed the include to the .scss file from the javascript entry point. We can dig in to the details and the repercussions of this change on video.

sherakama: Now I remember what stemmed all of this. 

I need to change the `$image-path` variable before compiling the sass and I cannot do that if it is included in the Javascript as I don't have an injection point before it gets included. I can get around this by allowing the decanter.js entry point output a CSS file that I don't use and just add another entry point to a SCSS file with the variable changes but that feels a bit messy. 

I like the idea of coupling the component css with the javascript but what do we do about the components that don't have Javascript?

JBCSU: Maybe it's better to talk & screen share, but these roll-up files in Scott's project are working:

`js/user.js`:
```
import "Decanter/js/decanter.js";
import "../scss/user.scss";
```

`scss/user.css`:
```
$image-path: '../../node_modules/decanter/core/src/img';
@import "decanter"; // decanter
```

It wasn't finding decanter's images before I overrode `$image-path`, so I'm highly confident that the override worked.

Anyway, yeah, let's talk & screen share and see if we can figure out how to get your project building.

BTW, this very override of `$image-path` in Scott's project made me wonder if we shouldn't have decanter-specific prefixes for variables (and mixins and placeholders?) like we do for class names. I can easily imagine a project that's built on top of decanter wanting it's own `$image-path` that doesn't affect where decanter looks for its images.

sherakama:
Ok, I got it going with the master branch but I do have a few questions regarding approach. I'll open up new PRs and issues.

sherakama:
Thanks for all the feedback.

webpack.config.js
sherakama: So you require building the decanter.css through the decanter.js file 🤔 Perhaps I just need to get used to this approach.

PR #379, opened 2019-04-05
add openNav, closeNav, openSubnav, closeSubnav events

JBCSU:
# READY FOR REVIEW

# Summary
In Scott's [event calendar project](http://events-dev.stanford.edu/), the **By Date** subnav contains a date picker. For accessibility reasons, he needs to place the focus on the date picker element when that subnav is opened. The cleanest way for him to do that is to add an event listener that responds whenever the appropriate subnav opens. He therefore needs an event to be triggered when a subnav opens. It seemed like it might also be useful to trigger events when a subnav closes, and when a mobile nav opens or closes.

# Needed By (Date)
Soon-ish, so Scott can respond to his accessibility issue.

# Urgency
Meh.

# Steps to Test

1. Pull the branch
1. `npm run build`
1. `npm run styleguide`
1. Visit your local dev instance in your browser and scroll down to the [Main Navigation](http://decanter.local/section-components.html#kssref-components-main-navigation) component
1. Open your console and paste the following commands
```
navs = document.querySelectorAll('.su-main-nav');
navs[1].addEventListener('openNav', (ev) => {console.log('mobile nav 1 OPENED');});
navs[1].addEventListener('closeNav', (ev) => {console.log('mobile nav 1 CLOSED');});
subnavs = document.querySelectorAll('.su-main-nav__item--parent');
subnavs[4].addEventListener('openSubnav', (ev) => {console.log('subnav 4 OPENED');});
subnavs[4].addEventListener('closeSubnav', (ev) => {console.log('subnav 4 CLOSED');});
subnavs[5].addEventListener('openSubnav', (ev) => {console.log('subnav 5 OPENED');});
subnavs[5].addEventListener('closeSubnav', (ev) => {console.log('subnav 5 CLOSED');});
```
6. With your console still open, try various ways of opening and closing  the **Useful Websites** and **Tools & Documentation** subnavs in the first example nav (**Default styling**). Notice the messages in the console.
6. Make your browser small enough so the example navs switch to mobile. With your console still open, try various ways of opening and closing  the first example nav (**Default styling**). Notice the messages in the console.

None. However, we need to figure out how Scott can get these changes. Right now he's fetching the `5.0.0` tag in the decanter repo to keep his environment stable. He probably shouldn't be pointing directly to `master`, as that's still volatile. Should we create a `5.0.1` tag? Or should we keep this branch alive after it's merged and have him point to this branch?

- @stocker

# See Also

### DIFF
--- a/core/src/js/components/main-nav/Nav.js
+++ b/core/src/js/components/main-nav/Nav.js
@@ -49,8 +49,8 @@ export default class Nav {
     );
 
     // add custom events to alert others when the mobile nav opens or closes
-    this.openEvent = createEvent( 'openNav' ); // dispatched in this.openMobileNav()
-    this.closeEvent = createEvent( 'closeNav' ); // dispatched in this.closeMobileNav()
+    this.openEvent = createEvent('openNav'); // dispatched in this.openMobileNav()
+    this.closeEvent = createEvent('closeNav'); // dispatched in this.closeMobileNav()
 
     elem.addEventListener('keydown', this);
 
@@ -280,7 +280,7 @@ export default class Nav {
    */
   closeMobileNav() {
     if (this.isExpanded()) {
-      this.setExpanded( 'false' );
+      this.setExpanded('false');
       this.toggle.innerText = this.toggleText;
       // alert others the mobile nav has closed
       this.elem.dispatchEvent(this.closeEvent);
--- a/core/src/js/components/main-nav/Nav.js
+++ b/core/src/js/components/main-nav/Nav.js
@@ -41,16 +41,20 @@ export default class Nav {
     this.toggle = elem.querySelector(elem.tagName + ' > button');
     this.toggleText = this.toggle ? this.toggle.innerText : '';
     this.items = [];
+    // Add custom events to alert others when the mobile nav
+    // opens or closes.
+    // this.openEvent is dispatched in this.openMobileNav().
+    this.openEvent = createEvent('openNav');
+    // this.closeEvent is dispatched in this.closeMobileNav().
+    this.closeEvent = createEvent('closeNav');
+
+    // Initialize items
     let items = elem.querySelectorAll(elem.tagName + ' > ul > li');
     items.forEach(
       item => {
         this.items.push(new NavItem(item, this));
       }
     );
-
-    // add custom events to alert others when the mobile nav opens or closes
-    this.openEvent = createEvent('openNav'); // dispatched in this.openMobileNav()
-    this.closeEvent = createEvent('closeNav'); // dispatched in this.closeMobileNav()
 
     elem.addEventListener('keydown', this);
 
@@ -214,8 +218,8 @@ export default class Nav {
    *                                relative to.
    */
   focusOn(link, currentItem = null) {
-    var currentIndex = null;
-    var lastIndex = null;
+    let currentIndex = null,
+      lastIndex = null;
     if (currentItem) {
       currentIndex = this.items.indexOf(currentItem);
       lastIndex = this.items.length - 1;
@@ -268,9 +272,10 @@ export default class Nav {
     this.setExpanded('true');
     this.toggle.innerText = 'Close';
     if (focusOnFirst) {
-      this.focusOn('first'); // Focus on the first top level link
+      // Focus on the first top level link.
+      this.focusOn('first');
     }
-    // alert others the mobile nav has opened
+    // Alert others the mobile nav has opened.
     this.elem.dispatchEvent(this.openEvent);
   }
 
@@ -282,7 +287,7 @@ export default class Nav {
     if (this.isExpanded()) {
       this.setExpanded('false');
       this.toggle.innerText = this.toggleText;
-      // alert others the mobile nav has closed
+      // Alert others the mobile nav has closed.
       this.elem.dispatchEvent(this.closeEvent);
     }
   }
--- a/core/src/js/components/main-nav/Nav.js
+++ b/core/src/js/components/main-nav/Nav.js
@@ -218,8 +218,8 @@ export default class Nav {
    *                                relative to.
    */
   focusOn(link, currentItem = null) {
-    let currentIndex = null,
-      lastIndex = null;
+    let currentIndex = null;
+    let lastIndex = null;
     if (currentItem) {
       currentIndex = this.items.indexOf(currentItem);
       lastIndex = this.items.length - 1;
--- a/core/src/js/components/main-nav/Nav.js
+++ b/core/src/js/components/main-nav/Nav.js
@@ -1,5 +1,6 @@
 import { closeAllMobileNavs } from './globals';
 import { isEsc, isSpace, isEnter } from "../../utilities/keyboard";
+import { createEvent } from '../../utilities/events';
 import NavItem from './NavItem';
 
 /**
@@ -46,6 +47,10 @@ export default class Nav {
         this.items.push(new NavItem(item, this));
       }
     );
+
+    // add custom events to alert others when the mobile nav opens or closes
+    this.openEvent = createEvent( 'openNav' ); // dispatched in this.openMobileNav()
+    this.closeEvent = createEvent( 'closeNav' ); // dispatched in this.closeMobileNav()
 
     elem.addEventListener('keydown', this);
 
@@ -265,6 +270,8 @@ export default class Nav {
     if (focusOnFirst) {
       this.focusOn('first'); // Focus on the first top level link
     }
+    // alert others the mobile nav has opened
+    this.elem.dispatchEvent(this.openEvent);
   }
 
   /**
@@ -272,8 +279,12 @@ export default class Nav {
    * initially.
    */
   closeMobileNav() {
-    this.setExpanded('false');
-    this.toggle.innerText = this.toggleText;
+    if (this.isExpanded()) {
+      this.setExpanded( 'false' );
+      this.toggle.innerText = this.toggleText;
+      // alert others the mobile nav has closed
+      this.elem.dispatchEvent(this.closeEvent);
+    }
   }
 
   // -------------------------------------------------------------------------
--- a/core/src/js/components/main-nav/NavItem.js
+++ b/core/src/js/components/main-nav/NavItem.js
@@ -32,13 +32,15 @@ export default class NavItem {
 
     if (this.isSubNavTrigger()) {
       this.subNav = new Nav(this);
+      // Add custom events to alert others when a subnav opens or closes.
+      // this.openEvent is dispatched in this.openSubNav().
+      this.openEvent = createEvent('openSubnav');
+      // this.closeEvent is dispatched in this.closeSubNav().
+      this.closeEvent = createEvent('closeSubnav');
+
       // Maintain global list of subnavs for closeAllSubNavs().
       theSubNavs.push(this);
       this.item.addEventListener('click', this);
-
-      // add custom events to alert others when a subnav opens or closes
-      this.openEvent = createEvent('openSubnav'); // dispatched in this.openSubNav()
-      this.closeEvent = createEvent('closeSubnav'); // dispatched in this.closeSubNav()
     }
   }
 
--- a/core/src/js/components/main-nav/NavItem.js
+++ b/core/src/js/components/main-nav/NavItem.js
@@ -1,7 +1,7 @@
-import { theSubNavs, closeAllSubNavs } from './globals';
-import { isHome, isEnd, isTab, isSpace, isEnter, isLeftArrow, isRightArrow, isUpArrow, isDownArrow } from "../../utilities/keyboard";
+import {theSubNavs, closeAllSubNavs} from './globals';
+import {isHome, isEnd, isTab, isSpace, isEnter, isLeftArrow, isRightArrow, isUpArrow, isDownArrow} from "../../utilities/keyboard";
 import Nav from './Nav';
-import { createEvent } from '../../utilities/events';
+import {createEvent} from '../../utilities/events';
 
 /**
  * Represent an item in a navigation menu. May be a direct link or a subnav
@@ -37,8 +37,8 @@ export default class NavItem {
       this.item.addEventListener('click', this);
 
       // add custom events to alert others when a subnav opens or closes
-      this.openEvent = createEvent( 'openSubnav' ); // dispatched in this.openSubNav()
-      this.closeEvent = createEvent( 'closeSubnav' ); // dispatched in this.closeSubNav()
+      this.openEvent = createEvent('openSubnav'); // dispatched in this.openSubNav()
+      this.closeEvent = createEvent('closeSubnav'); // dispatched in this.closeSubNav()
     }
   }
 
@@ -130,7 +130,7 @@ export default class NavItem {
       if (focusOnFirst) {
         this.subNav.focusOn('first');
       }
-      this.item.dispatchEvent( this.openEvent );
+      this.item.dispatchEvent(this.openEvent);
     }
   }
 
@@ -146,12 +146,12 @@ export default class NavItem {
   closeSubNav(focusOnTrigger = false) {
     if (this.isSubNavTrigger()) {
       if (this.isExpanded()) {
-        this.item.classList.remove( 'su-main-nav__item--expanded' );
-        this.setExpanded( 'false' );
-        if ( focusOnTrigger ) {
+        this.item.classList.remove('su-main-nav__item--expanded');
+        this.setExpanded('false');
+        if (focusOnTrigger) {
           this.link.focus();
         }
-        this.item.dispatchEvent( this.closeEvent );
+        this.item.dispatchEvent(this.closeEvent);
       }
     }
     else if (this.isSubNavItem()) {
--- a/core/src/js/components/main-nav/NavItem.js
+++ b/core/src/js/components/main-nav/NavItem.js
@@ -1,7 +1,7 @@
 import { theSubNavs, closeAllSubNavs } from './globals';
 import { isHome, isEnd, isTab, isSpace, isEnter, isLeftArrow, isRightArrow, isUpArrow, isDownArrow } from "../../utilities/keyboard";
 import Nav from './Nav';
-import '../../utilities/keyboard';
+import { createEvent } from '../../utilities/events';
 
 /**
  * Represent an item in a navigation menu. May be a direct link or a subnav
@@ -35,6 +35,10 @@ export default class NavItem {
       // Maintain global list of subnavs for closeAllSubNavs().
       theSubNavs.push(this);
       this.item.addEventListener('click', this);
+
+      // add custom events to alert others when a subnav opens or closes
+      this.openEvent = createEvent( 'openSubnav' ); // dispatched in this.openSubNav()
+      this.closeEvent = createEvent( 'closeSubnav' ); // dispatched in this.closeSubNav()
     }
   }
 
@@ -126,6 +130,7 @@ export default class NavItem {
       if (focusOnFirst) {
         this.subNav.focusOn('first');
       }
+      this.item.dispatchEvent( this.openEvent );
     }
   }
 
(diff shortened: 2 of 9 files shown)


josephgknox:
@JBCSU are any of the CC issues ones that you think we should correct before merging? Does out linter need to be updated to ignore such flags (looks like 7 of the 8 issues are the same)?

JBCSU:
> @JBCSU are any of the CC issues ones that you think we should correct before merging? Does out linter need to be updated to ignore such flags (looks like 7 of the 8 issues are the same)?

7 of them are effectively complaining that I commented my code: the complaint is that the total line length, including my comments, is >80 characters. IMO we should change our linter to either not count comments in character counts, or allow longer lines. Or, IMO, both. 80 characters was the length of a punched card. It's 2019 and we're all using IDE's. I think we could go to 100, 120, maybe longer? And I think comment characters should be ignored, or perhaps even subtracted from overall line length! Comments are Good Things.

The other one refers to the "cognitive complexity" I introduced by only collapsing subnavs if they're actually open. Previously I "collapsed" them unconditionally because it was just DOM tweaking, which is very cheap. But now that I'm actually firing an event, which is much more costly, I only want to fire the event if the subnav actually collapsed. That's a 14 line function that IMO is eminently readable. I am 100% confident that anyone looking at that function could understand it perfectly well. So no, I don't think that CC complaint needs to be addressed either.

But those are my opinions. You guys are my reviewers, so you tell me.

josephgknox:
As always, thanks so much for the detailed reply! This makes sense to me, and I am behind each of your opinions here. My vote would be to update our linters to allow for longer lines as well as not count comments. Let's see if others have thoughts, too.

core/src/js/utilities/events.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Create an event with the specified name in a browser-agnostic way.
+ * @param {string} eventName - the name of the event
```
sherakama: This parameter is not {string} bound but rather mixed as it accepts strings or functions.

core/src/js/utilities/events.js
```diff
@@ -0,0 +1,18 @@
...
+ * @return {Event} - instance of event which can be dispatched / listened for
+ */
+export const createEvent = (eventName = '') => {
```
sherakama: Do you need the default string? I'd vote no as it is not useful to have an empty string passed in.

core/src/js/utilities/events.js
```diff
@@ -0,0 +1,18 @@
...
+  if (typeof eventName !== 'string' || eventName.length <= 0) {
+    return null;
```
sherakama: Do we want this to quietly fail or do we want to throw an exception here?

core/src/js/components/main-nav/Nav.js
```diff
@@ -47,6 +48,10 @@ export default class Nav {
       }
     );
 
+    // add custom events to alert others when the mobile nav opens or closes
```
sherakama:
```suggestion
    // Add custom events to alert others when the mobile nav opens or closes.
```

sherakama: Can we move this new block up to line 35/36 so that the instance property assignments are all together before the logic statements?

core/src/js/components/main-nav/Nav.js
```diff
@@ -265,15 +270,21 @@ export default class Nav {
     if (focusOnFirst) {
       this.focusOn('first'); // Focus on the first top level link
     }
+    // alert others the mobile nav has opened
```
sherakama:
```suggestion
    // Alert others the mobile nav has opened.
```

core/src/js/components/main-nav/Nav.js
```diff
@@ -265,15 +270,21 @@ export default class Nav {
...
   }
 
   /**
    * Mark this mobile closed, and restore the button text to what it was
    * initially.
    */
   closeMobileNav() {
-    this.setExpanded('false');
-    this.toggle.innerText = this.toggleText;
+    if (this.isExpanded()) {
```
sherakama: Why do we need to check for expanded here? Does it matter?

core/src/js/components/main-nav/Nav.js
```diff
@@ -265,15 +270,21 @@ export default class Nav {
...
+      this.setExpanded('false');
+      this.toggle.innerText = this.toggleText;
+      // alert others the mobile nav has closed
```
sherakama:
```suggestion
      // Alert others the mobile nav has closed.
```

core/src/js/components/main-nav/NavItem.js
```diff
@@ -35,6 +35,10 @@ export default class NavItem {
       // Maintain global list of subnavs for closeAllSubNavs().
       theSubNavs.push(this);
       this.item.addEventListener('click', this);
+
+      // add custom events to alert others when a subnav opens or closes
```
sherakama:
```suggestion
      // Add custom events to alert others when a subnav opens or closes.
```

core/src/js/components/main-nav/NavItem.js
```diff
@@ -140,10 +145,13 @@ export default class NavItem {
    */
   closeSubNav(focusOnTrigger = false) {
     if (this.isSubNavTrigger()) {
-      this.item.classList.remove('su-main-nav__item--expanded');
-      this.setExpanded('false');
-      if (focusOnTrigger) {
-        this.link.focus();
+      if (this.isExpanded()) {
```
sherakama: Why do we need to check if the item is expanded here? Is it destructive to execute the close items if it wasn't open?

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
   const navs = document.querySelectorAll('.' + navClass);
 
-  // ---------------------------------------------------------------------------
-  // Execution Code.
-  // ---------------------------------------------------------------------------
-
   // Process each nav.
   let firstZindex;
   navs.forEach((nav, index) => {
+    // remove the class that formats the nav for browsers with javascript disabled
```
sherakama:
```suggestion
    // Remove the class that formats the nav for browsers with javascript disabled.
```

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
+    nav.classList.remove('no-js');
```
sherakama: This is happening in the template file. Can we remove that javascript code from there now?

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
+
+    // create an instance of Nav, which in turn creates appropriate instances of NavItem
```
sherakama:
```suggestion
    // Create an instance of Nav, which in turn creates appropriate instances of NavItem.
```

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
     const theNav = new Nav(nav);
+
+    // remember the nav for closeAllMobileNavs()
```
sherakama:
```suggestion
    // Remember the nav for closeAllMobileNavs().
```

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
     theNavs.push(theNav);
 
+    // manage zindexes in case there are multiple navs near enough for subnavs to overlap
```
sherakama:
```suggestion
    // Manage zindexes in case there are multiple navs near enough for subnavs to overlap
```

core/src/js/components/main-nav/main-nav.js
```diff
@@ -6,22 +6,23 @@ document.addEventListener('DOMContentLoaded', event => {
...
+    // extremely unlikely, but it happens in the styleguide
```
sherakama:
```suggestion
    // extremely unlikely, but it happens in the styleguide.
```

core/src/js/utilities/events.js
```diff
@@ -0,0 +1,18 @@
...
+  }
+  if (typeof Event == 'function') { // modern browsers
```
sherakama:
```suggestion
  // Modern browsers.
  if (typeof Event == 'function') {
```

core/src/js/utilities/events.js
```diff
@@ -0,0 +1,18 @@
...
+    return new Event(eventName);
+  }
+  else { // IE :(
```
sherakama:
```suggestion
  // IE :(
  else {
```

sherakama:
@JBCSU the 80 char limit is pretty common across many projects. Keeping the comments above the item in which they refer to and at 80 characters keep formatting consistent. 

https://nickjanetakis.com/blog/80-characters-per-line-is-a-standard-worth-sticking-to-even-today
https://hackernoon.com/does-column-width-of-80-make-sense-in-2018-50c161fbdcf6

Seems like there is some movement to increase this line limit to upwards of 120. 
Looks like airbnb's standards are at 100: https://github.com/airbnb/javascript/pull/639/files
Bootstrap disabled theirs.

I think we should keep a max character width in our code and I am open to increasing it. Should it be 100 or 120?

core/src/js/utilities/events.js
JBCSU: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event), the first parameter to `new Event()` is a DOMString, and "[DOMString is mapped directly to a String](https://developer.mozilla.org/en-US/docs/Web/API/DOMString)." Also, my wrapper function explicitly test for string, and fails if it's not a (non-empty) string.

core/src/js/utilities/events.js
JBCSU: Normally I like to defend against errors, but in this case I guess a compile time error is better than a runtime error. I'll remove the default.

core/src/js/components/main-nav/Nav.js
JBCSU: The only logic that's above it is initializing `this.items`. I like to keep the initializing of properties next to the declarations. That way I only have to look in 1 place when I need to know what the property is.

core/src/js/components/main-nav/Nav.js
JBCSU: I really don't like the sentence-like construction of comments. Can we agree to disagree on that comment style?

core/src/js/components/main-nav/Nav.js
JBCSU: I didn't have it before. But now that we're firing an event when the nav closes, yes, I think it matters:

- Firing an event is much more expensive than DOM tweaking. I don't think we should incur that cost unless we need to.
- If someone is listening for the `closeNav` event, I think they'll only want to be notified if the nav actually closed. Right now we call `closeMobileNav()` anytime anyone clicks outside of a nav. We don't want to fire `closeNav` every time that happens.

core/src/js/components/main-nav/NavItem.js
JBCSU: See comment above about `closeMobileNav()`. Same reasoning applies here.

core/src/js/components/main-nav/main-nav.js
JBCSU: I thought I still saw a brief FOUC when I removed it from the template file, but it seems ok now. I'm happy to remove it.

core/src/js/components/main-nav/main-nav.js
JBCSU: If I were totally unfamiliar with Decanter's nav, but reading through the code in anticipation of using it in a project, I would be deeply puzzled by that zIndex loop. I would be reassured by knowing that the problem it's trying to solve is in fact extremely unlikely in production.

core/src/js/utilities/events.js
JBCSU: I think there's room for both styles of commenting. I like comments on their own line when they're introducing a distinct chunk of code. That style of commenting helps me chunk up code mentally, in part because it interrupts my reading of the code. I like comments at the end of the line when they're just asides to help clarify a particular line or two of code. They're off to the side, so they don't interrupt my reading of the code.

JBCSU:
> I think we should keep a max character width in our code and I am open to increasing it. Should it be 100 or 120?

If trailing comments count, then I vote for 120. If only code counts, then 100 seems fine.

core/src/js/components/main-nav/Nav.js
sherakama: In javascript variables are hoisted. See: https://www.w3schools.com/js/js_hoisting.asp

It is best practices to always declare your variables and properties at the top. 

> To avoid bugs, always declare all variables at the beginning of every scope.
> Since this is how JavaScript interprets the code, it is always a good rule.

```
    this.elem = elem;
    this.topNav = this.getTopNav();
    
    // Add custom events to alert others when the mobile nav opens or closes
    this.openEvent = createEvent('openNav'); // dispatched in this.openMobileNav()
    this.closeEvent = createEvent('closeNav'); // dispatched in this.closeMobileNav()

    this.toggle = elem.querySelector(elem.tagName + ' > button');
    this.toggleText = this.toggle ? this.toggle.innerText : '';
    this.items = [];

    // If this is a subnav, we need the correpsonding HTMLElement for
    // .querySelector()
    if (elem instanceof NavItem) {
      elem = elem.item;
    }
   
    let items = elem.querySelectorAll(elem.tagName + ' > ul > li');
    items.forEach(
      item => {
        this.items.push(new NavItem(item, this));
      }
    );
```

core/src/js/components/main-nav/Nav.js
sherakama: This is a convention that has been hammered into me by multiple other projects. I don't see anything that would make this a requirement in this project. Checking a few of the most popular projects on npm.org, I see the following:

lodash: Uses sentence structure
express: Doesn't use sentence structure
react: Capital start but no period.
chalk: Mixed, no format.
request: Capital start but no period.
commander: Capital start but no period.
Moment: Uses sentence structure
debug: Uses capitals but no period.

How about we split the difference? Comments must start with a capital letter but no punctuation is required at the end? That seems to be the common pattern.

core/src/js/components/main-nav/main-nav.js
sherakama: If you feel strongly about this comment then I am fine to leave it along. My experience in working with open source projects is that I shouldn't try to guess what crazy scenarios another developer may have. Forces unknown can lead to wacky scenarios.

core/src/js/utilities/events.js
sherakama: Checking the top packages on npmjs.org

lodash: Seems to be above but mostly just has function comments
request: mixed, some above and some inline
chalk: mixed, some above and some inline
react: above
debug: above
moment: mixed, some above and some inline
express: mixed, commonly inline.

Checking wordpress standards it appears they use a modified jQuery convention:
- https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/
- https://contribute.jquery.org/style-guide/js/#comments

> Comments are always preceded by a blank line. Comments start with a capital first letter, but don't require a period at the end, unless you're writing full sentences. There must be a single space between the comment token and the comment text.

I try to stick to the conventions of the project I intend to be working in as it keeps everything similar but it looks like it is fairly mixed in some of the top npm packages.

core/src/js/components/main-nav/Nav.js
sherakama: Sounds like a good idea to me. Out of curiosity, what did you use to measure the expense of the event call?

core/src/js/components/main-nav/NavItem.js
sherakama: Thanks, GTG here.

sherakama:
Ok, 120 it is. 

Just as an FYI, it looks like Wordpress uses the 80 character limit. https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#commenting

core/src/js/utilities/events.js
sherakama: Oh, I totally read the if statements wrong. I thought the second check where the if statement is checking to see if Event is a function was on the input parameter. You are 100% right. Apologies for this comment.

core/src/js/components/main-nav/Nav.js
JBCSU: > Out of curiosity, what did you use to measure the expense of the event call?

Intuition. Plus, I watched my event handlers getting called **WAY** more often than they needed to be. :)

core/src/js/utilities/events.js
JBCSU: > I try to stick to the conventions of the project I intend to be working in as it keeps everything similar but it looks like it is fairly mixed in some of the top npm packages.

I do try to stick to the conventions of the project I’m working in. But since this is the very first Javascript in the project, I thought there might be some flexibility around what those conventions are.

Also, to my mind, conventions are not ends in themselves. They’re there to make reading and understanding the code easier. I honestly cannot imagine a scenario where a comment beginning with a lower case letter vs. an upper case letter would affect anyone’s ability to read and understand the code. I therefore don’t think it’s in anyone’s best interests to bother having a convention about that. I accept that some people do make conventions about that, but to my mind that’s not a productive use of anyone’s time.

Of course discussing comment style at this length is also not a productive use of our time. I’ve changed the code per your requests.

core/src/js/components/main-nav/Nav.js
JBCSU: I updated the constructor per your request.

core/src/js/components/main-nav/main-nav.js
JBCSU: I modified the comment to say "Rare, but..."

core/src/js/utilities/events.js
sherakama: I'd argue that the very first of something is the best time to start a convention. Instead of letting something snowball into work later, nip it early and carry on. Starting early is also a benefit to the early contributors as they get the biggest say into what the convention should be.

I agree that capitalization and punctuation of comments is not a make or break scenario for readability but rather, my concerns stem from having consistency in style. Consistency plays a big part into readability and is good for the overall health of the project. Many other people much smarter than myself have had long discussions on the merits of each style and have created unified standards for their projects. I'd like to stay in line with the projects we intend to use this code in and when there isn't one, pick from other popular projects. 

My goal of these reviews is not to have long pedantic conversations about style, but rather, to call out and pick a standard when I see a difference. Please bear with this process as it gets quicker as we go.

core/src/templates/components/main-nav/main-nav.twig
```diff
@@ -65,8 +65,3 @@
     %}
   {% endif %}
 </nav>
-<script>
```
sherakama: woo hoo. Thanks for this!.

core/src/js/utilities/events.js
JBCSU: I trust that we both have the best interests of the project at heart. But I think the comment thread of what was meant to be a minor PR is likely not the best forum for proposing, discussing or determining these kinds of coding standards, especially when there's clearly some disagreement. If we want to invest time in coming to a shared consensus on standards, I'm happy to discuss it in a call, either 1-1 or in a regular meeting. Alternatively, I did propose including a coding standards issue in our upcoming Decanterpalooza, for just this reason. But in the meantime, for the sake of this PR, I believe I have made all the changes you requested. I'm happy to shake hands, reaffirm our shared commitment to writing excellent code that benefits Stanford, and go forth and conquer more components! :)

core/src/templates/components/main-nav/main-nav.twig
JBCSU: Fer shur. That needed to go!

sherakama:
Thank you for the learning opportunities and valuable feedback.

core/src/js/utilities/events.js
sherakama: That's a really good point. We could speed things up by spending some time hashing out the lint files. +1 to getting to those tickets sooner than later.

PR #380, opened 2019-04-09
Fix main nav color variants and work with new linting rules

yvonnetangsu:
# READY FOR REVIEW

# Summary
- I noticed that the color variants (e.g., the --dark version on the decanter website) aren't working properly, e.g., when you click on a parent link item (with downward caret) on the dark variant (the one on decanter website), the caret disappears. This fix should make them behave as intended. Also, tweak the code to work with new linting rules.
- Add drop shadow for MD breakpoint mobile menu (since it doesn't span full width like in xs/sm breakpoints) which helps distinguish menu from the background and it's consistent with 1st level drop down menus on desktop version.

# Needed By (Date)
- N/A

# Urgency
- Sooner the better because the version on the website is acting strange.

# Steps to Test

1. Pull this branch and run `npm run build`
2. Check that the --light and --dark variants are working correctly. Check that at md breakpoint, there is drop shadow around the menu when it's open.
3. Run `npm run sasslint` to check that there are no warnings for the main nav component

- Decanter

### DIFF
--- a/core/src/scss/components/main-nav/_main-nav--dark.scss
+++ b/core/src/scss/components/main-nav/_main-nav--dark.scss
@@ -1,99 +1,51 @@
 @charset "UTF-8";
 
-.su-main-nav--dark {
-  @include grid-media-max('md') {
-    .su-main-nav__toggle {
-      color: $color-white;
+.su-main-nav {
+  &.su-main-nav--dark {
+    @include grid-media-max('md') {
+      .su-main-nav__toggle {
+        color: $color-white;
 
-      &:hover,
-      &:focus {
-        &::before {
-          background-color: $color-white;
+        &:hover,
+        &:focus {
+          &::before {
+            background-color: $color-white;
+          }
         }
-      }
 
-      &:active {
-        &::before {
-          background-color: $color-vivid-red;
-        }
-      }
-
-      &::after {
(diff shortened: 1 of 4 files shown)


core/src/scss/components/main-nav/_main-nav--dark.scss
```diff
@@ -1,125 +1,127 @@
...
-      &:focus {
-        &::before {
-          background-color: $color-white;
-        }
-      }
-
-      &:active {
-        &::before {
-          background-color: $color-vivid-red;
+.su-main-nav {
```
sherakama: It's too bad you had to do that. I would prefer to see that we don't double up on the modifier classes when possible.   Using the CSS cascade to have modifiers come after the main styles can be enough to have more specificity. I would challenge you to try to take another look and see if you can refactor this to win out over the main styles without the double class selector.

yvonnetangsu: Yes, I agree with you! Sorry I didn't see this comment before I merged. I would definitely go back and fine tune it when I have a chance.

sherakama: You are fast on the trigger 🔫! Yeah, this comment doesn't block the merge but it is something that could use attention down the line and not a pattern I want to see too often.

Issue #414, opened 2019-05-01
SR only text helper causes overflow


yvonnetangsu:
**Describe the bug**
When used on some sites, the .su-sr-only-text helper causes overflowing and takes up height (shows empty space).

**Expected behavior**
The .su-sr-only-text shouldn't take up any space and should only be invisible and not change the page's appearance in any way.

**Screenshots**

![Screen Shot 2019-05-01 at 2 58 41 PM](https://user-images.githubusercontent.com/42749717/57045231-c406a080-6c21-11e9-8acd-2e4b6b4e423d.png)

The blue focus ring shows the empty space caused by the .su-sr-only-text (skip link h2) and also the .su-sr-only-text after "Read me" is still showing and overflowing.

kgcreative:
I seem to recall there being a .sr-only-text which is good for in-line elements, vs either .sr-only-element, or .hide-visually (I forget which one), which hides block-level elements.

yvonnetangsu:
> I seem to recall there being a .sr-only-text which is good for in-line elements, vs either .sr-only-element, or .hide-visually (I forget which one), which hides block-level elements.

Yup, just opened a PR to use .hide-visually instead 👍

kgcreative:
Good detective work! 😎 also, hi Yvonne! 👋

yvonnetangsu:
Great seeing you here, @kgcreative ! 😄

PR #415, opened 2019-05-01
Fix skiplink target so it doesn't take up space and cause overflow

yvonnetangsu:
# READY FOR REVIEW

# Summary
- The skiplink target was using the `.su-sr-only-text` class which uses the Bourbon `@hide-text` mixin which takes up space and causes overflow. This PR creates new class `.su-skiplinks__target` which uses the `@hide-visually` mixin instead which has the style code we need. The skip link target now has proper styles (same as current best practice).
- I didn't just use the `.su-sr-only-element` class on the skip link target directly because .su-sr-only-element unhides on focus.
- Add documentation for `.su-sr-only-text` - it's only good for cases where e.g., hide the SR text in the site-search component magnifying glass button/icon.

# Needed By (Date)
- ASAP

# Urgency
- Pretty urgent

# Steps to Test
1. Pull this branch and run `npm run build`.
2. Look at the skiplink target in the skiplinks component. It should now now take up any spaces.

- Decanter

- #414

### DIFF
--- a/core/src/scss/components/skiplinks/_skiplinks.scss
+++ b/core/src/scss/components/skiplinks/_skiplinks.scss
@@ -19,7 +19,7 @@
 //   <a href="#main-content" class="su-skiplinks">Skip to main content</a>
 // </header>
 //  <main>
-//   <h2 id="main-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
+//   <h2 id="main-content" class="su-skiplinks__target" tabindex="-1">Main Content</h2>
 // </main>
 //
 //
@@ -28,4 +28,8 @@
 //
 .su-skiplinks {
   @include skiplinks;
+}
+
+.su-skiplinks__target {
+  @include hide-visually;
 }
--- a/core/dist/css/decanter.css
+++ b/core/dist/css/decanter.css
@@ -8768,5 +8768,17 @@ a {
     width: auto;
     z-index: 11222; }
 
+.su-skiplinks__target {
+  border: 0;
+  clip: rect(1px, 1px, 1px, 1px);
+  -webkit-clip-path: inset(100%);
+          clip-path: inset(100%);
+  height: 1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px; }
+
 
 /*# sourceMappingURL=decanter.css.map*/
--- a/core/src/scss/core/helpers/_sr-only-text.scss
+++ b/core/src/scss/core/helpers/_sr-only-text.scss
@@ -3,7 +3,9 @@
 //
 // Screen Reader Only Text
 //
-// Hides the text in an element, commonly used to show an image instead.
+// Use this helper only for text that describes an image but should be hidden visually. For other use cases where
+// text needs to be hidden visually (e.g., skip link targets), use the .su-sr-only-element helper class instead.
+//
 // Some elements will need block-level styles applied.
 //
 // Style guide: Core.Helpers.sr-only
--- a/core/src/templates/components/skiplinks/target.twig
+++ b/core/src/templates/components/skiplinks/target.twig
@@ -13,4 +13,4 @@
  *   skiplink - defaults to 'Main Content'
  */
 #}
-<h2 {{ attributes }} id="{{ target_id|default('su-main-content-1891') }}" class="su-sr-only-text {{ modifier_class }}" tabindex="-1">{{ target_text|default('Main Content') }}</h2>
+<h2 {{ attributes }} id="{{ target_id|default('su-main-content-1891') }}" class="su-skiplinks__target {{ modifier_class }}" tabindex="-1">{{ target_text|default('Main Content') }}</h2>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "decanter",
-  "version": "5.0.1",
+  "version": "5.1.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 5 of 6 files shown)


core/src/scss/components/skiplinks/_skiplinks.scss
```diff
@@ -19,7 +19,7 @@
 //   <a href="#main-content" class="su-skiplinks">Skip to main content</a>
 // </header>
 //  <main>
-//   <h2 id="main-content" class="su-sr-only-text" tabindex="-1">Main Content</h2>
+//   <h2 id="main-content" class="su-skiplinks__target" tabindex="-1">Main Content</h2>
```
JBCSU: Why do we need a new class? Why can't we just fix `.su-sr-only-text`?

If we do need a new class, I'd prefer to not use a name that implies you can only use the class on one item, e.g. skiplinks. I would expect the class to be useful for anything that needs to be hidden, e.g. the supplemental text in **Read more** links. Is there some reason the skip link target need to be treated specially?

yvonnetangsu: .su-sr-only-text uses the Bourbon mixin @hide-text which is good for some use cases, eg, for the SR only text in site search magnifying glass icon so I want to preserve that.

We currently have 2 existing SR helper classes (the above mentioned .su-sr-only-text, and .su-sr-only-element) and neither of them function the way we want the skiplink target to function. One way we could make this solution more general would be to rethink .su-sr-only-element. Currently, it unhides when it's on-focus. I don't believe any components in Decanter uses it, but this class might be used by some other sites that use Decanter? This might be something that needs a discussion.

JBCSU: Are you saying your implementation, `@hide-visually`, wouldn't work for the search icon?

yvonnetangsu: Yeah, I believe that's what I tried initially and it didn't work.

JBCSU: Ok, so we have a few issues here:

Apparently the point of `.su-sr-only-text` is to hide just the textual content of an element, but render everything else, e.g. background, border, padding, .... `.su-sr-only-element` is to hide the entire element, including background, border, padding, .... Applying `.su-sr-only-element` to the search icon hides the icon as well as the helper text, because the icon is a background image for the helper text.

In testing this, I found the first issue, which is that the search icon's helper text is never read aloud by voice over - the `aria-label="Search"` is read, regardless of what text is provided for the `<button>`. So that helper text is completely ignored, at least by voice over - we can remove it entirely and therefore we can remove `.su-sr-only-text` from the button.

If the target of the skip link is, as we recommend, an `<h2>` whose sole purpose is to announce to screen-readers "Main content", then we want to hide the entire element. So `.su-sr-only-element` seems appropriate. So now we need to decide what should happen when focus lands on the sr-only-element.

I propose that `.su-sr-only-element` does not reveal on focus, and that we add a variant, `.su-sr-only-element--reveal-on-focus`, that does.

That said, I think the target of the skip link, which is the aforementioned `<h2>`, actually should reveal on focus. When I'm testing keyboard nav and I hit the skip link, I'm left wondering where my focus went. I have to keep tabbing to figure out where I'm at. I think if the `<h2>Main Content</h2>` revealed itself it would be as useful to keyboard navigators as it is to screen readers. Jiatyan agrees that that would be helpful.

yvonnetangsu: I'm curious about the aria-label/button text issue also. If Level Access can confirm that it is ok to remove the button text entirely then we can certainly just do that.

I'm open to the idea of skip link targets reveals itself on focus. I also agree that our default `.su-sr-only-element` should not reveal on focus and I like the idea of a variant that does reveal itself on focus.

Other Decanter people - any thoughts? 😃

sherakama: Might as well use the bourbon mixins for this. As the target is its own element I would suggest changing the name to `su-skiplinks-target` instead of the BEM child as it really isn't nested. If you want to reveal the element on focus you can use the unhide paramater as suggested in the bourbon documentation.

```
/// @example scss
///   .su-skiplinks-target {
///     @include hide-visually;
///
///     &:active,
///     &:focus {
///       @include hide-visually("unhide");
///     }
///   }
```

sherakama: As this is pretty much all new it doesn't break compatibility and we can safely merge and go on.

sherakama:
Original issue here: https://github.com/SU-SWS/decanter/issues/414

yvonnetangsu:
> As the target is its own element I would suggest changing the name to `su-skiplinks-target` instead of the BEM child as it really isn't nested.
> 
> Other than that this looks good to me.

Thanks for the review, @sherakama ! I had a discussion with JB last week and she's not keen on the idea of adding a new class name so instead she'd rather we tweak the default `.su-sr-only-element` class and remove the unhide on focus behavior, and adding a variant to `.su-sr-only-element` class that does unhide on focus. @JBCSU - comments? Would that work, @sherakama?

yvonnetangsu:
@sherakama , @JBCSU , I removed the su-skiplinks__target class, used the su-sr-only-element class instead and changed the su-sr-only-element helper class so it doesn't unhide on focus. One thing I didn't do is putting unhide on focus in the skiplink target - I remember @JBCSU , Scott and I had a discussion and he preferred that skiplink targets do not unhide on focus, since that's the conventional thing out there for skiplinks and he didn't want to confuse keyboard users?

PR #416, opened 2019-05-17
Move PR template

sherakama:
* Moves the PR template so that it is the default PR body when creating a new PR.
* The old place for the template required URL parameters.


yvonnetangsu:
Sounds good to me!

Issue #386, opened 2019-04-16
Refactor main nav's default, light, and dark variants to use maps


JBCSU:
On the [new events calendar](events-dev.stanford.edu), Scott needs to override some of the colors in the nav, including hover states, expanded states, and collapsed states. That is currently very difficult because the nav's .scss uses deeply nested selectors that refer to colors directly from the color palette. He can't just change the value of, for example, `$color-black` before he includes `_main-nav--light.scss`, because that would change black everywhere. His only alternative currently is to copy and paste the light variant's entire scss and replace the color names.

I would like for the colors of the default nav, as well as the light and dark variants, to be specified in a scss map, similar to how theme colors are defined in homesite. That would allow someone to specify a custom map to get customized colors for their nav without having to either change a core color or completely replicate the scss in their project. It would probably also reduce the redundancy in Decanter's scss, as the same scss could be for all the variants - we just re-use it with different maps.

Note that currently we use different svg's to get different colors for the collapsed / expanded indicators. Hopefully the maps can include the path to the correct svg, so the colors of those controls can also be customized.

sherakama:
I like where you are going with this idea and this definitely ties in with the global levers issue we have.

As a quick aside, you can color SVGs using this one neat css trick...
https://css-tricks.com/almanac/properties/f/fill/

JBCSU:
That `fill` property is a neat trick to know to recolor foreground svg's. Sadly, it doesn't seem to work for the nav icons, presumably because they're background images. :( But thanks for the tip!

yvonnetangsu:
It might be worth it for us to do more research into how best to use svg icons. Perhaps use image instead as content CSS property, or image-mask (which allows you to change the color of the icon with the change of background-color).

sherakama:
https://css-tricks.com/change-color-of-svg-on-hover/

yvonnetangsu:
@sherakama, yeah, the mask option seems pretty promising. I also looked at the filter one initially, but it doesn't seem to work on IE (maybe even Edge) - that calculator for target color is handy though!

yvonnetangsu:
Done

PR #420, opened 2019-05-31
Set up color maps for Decanter and refactor main nav colors to refer to color maps

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Set up color maps for Decanter - a main one `$core-colors` and one for main-nav `$main-nav-colors`. There is already an existing one called `$alert-colors`.
- Add a function `color` to get colors from color maps.
- Refactor main-nav and a few other components to pull colors from the color map
- Also included issue 422 - changed all reference to $color-dark-red to $color-bright-red for accessibility

# Needed By (Date)
- Whenever

# Urgency
- Not urgent

# Steps to Test

1. Pull this branch
2. See that the main nav, buttons, brand bar, logo, global footer, alerts, links and all variants still have the correct colors.
3. Look at the new color maps in utilities/variables and the function `color`.
4. See that all places in main-nav with $color-dark-red have been changed to $color-bright-red.

- Decanter

- https://github.com/SU-SWS/decanter/issues/386
- https://github.com/SU-SWS/decanter/issues/420
- https://github.com/SU-SWS/decanter/issues/422


core/src/scss/components/alert/_alert--error.scss
```diff
@@ -2,7 +2,7 @@
 
 // Error alert.
 .su-alert--error {
-  background-color: map-get($alert-colors, 'error');
+  background-color: color(error, $alert-colors);
```
sherakama: Oooh, I like this. Super clean.

core/src/scss/components/main-nav/_main-nav--dark.scss
```diff
@@ -1,128 +1,137 @@
...
-    @include grid-media-max('md') {
-      .su-main-nav__toggle {
-        color: $color-white;
-
-        &:hover,
-        &:focus {
-          &::before {
-            background-color: $color-white;
-          }
+.su-main-nav--dark {
```
sherakama: Yes! less nesting FTW.

core/src/scss/components/main-nav/_main-nav--dark.scss
```diff
@@ -1,128 +1,137 @@
...
+  @include grid-media-max('md') {
+    .su-main-nav__toggle {
+      color: color(link, $main-nav-colors, dark);
+
+      &:hover,
+      &:focus {
+        &::before {
+          background-color: color(border--hover, $main-nav-colors, dark);
```
sherakama: This is a little less clear than the two parameter function. Instead of having nested themes, why wouldn't you just pass in a different color map? eg:

`background-color: color(border--hover, $main-nav-colors-dark);`

core/src/scss/utilities/functions/_color.scss
```diff
@@ -0,0 +1,53 @@
...
+// If all 3 are specified, it returns a color from the specified theme of the color map.
+//
+// $color - String - Color variable name from a color map, e.g., brand, link--hover
+// $colormap - Color palette in the form of a SASS map, e.g., $main-nav-colors, $alert-colors. Default is $core-colors.
+// $theme - String - Name of the theme within a color map, e.g., dark, light. Default is default.
+//
+// Style guide: Functions.Color
+//
+
+@function color($color, $colormap: $core-colors, $theme: default) {
```
sherakama: I think the third parameter (theme) could be removed in favor of simply passing in a different colormap.

sherakama: Also, this is a great interface for color changes. YES THANK YOU!

yvonnetangsu:
> I like this, I like this a lot. My only suggestion is to simplify the color function by removing the third parameter but I am open for influence or feedback from others on this.

Awesome 😄 Actually you're right. I'm not sure there's any real reason to use the nested maps. @JBCSU , are you ok with @sherakama 's suggestion to not have nested maps?

sherakama:
We should only be rendering one 'theme' at a time. To do that, we will need some sort of toggle and that will likely be in some sort of variable changes. If I were to create themes today I would create a new partial file with all of the variable changes and include it in my rendering of css. We could package themes in this project by adding those partial files as new entry points in webpack but I think I would rather provide documentation and an examples file instead of having to support multiple themes per component. We could unofficially share those theme partials across our projects through gists or something else.

yvonnetangsu:
@sherakama @JBCSU I've removed the nesting in the color maps and changed the function @color to get rid of the $theme parameters. Also, I've changed all reference to $color-dark-red to $color-bright-red per @kerri-augenstein 's suggestion (https://github.com/SU-SWS/decanter/issues/422). Could you please give it another look? Thanks!

core/src/scss/components/main-nav/_main-nav--dark.scss
```diff
@@ -1,128 +1,139 @@
...
-      .su-main-nav__toggle {
-        color: $color-white;
-
-        &:hover,
-        &:focus {
-          &::before {
-            background-color: $color-white;
-          }
+.su-main-nav--dark {
+  $_color-map: $main-nav-dark-colors;
```
sherakama: Just in case this variable name is used somewhere else, can you prefix it with the component name. 

eg: `$main-nav-color-map`

core/src/scss/components/main-nav/_main-nav--light.scss
```diff
@@ -1,189 +1,233 @@
 @charset "UTF-8";
 
-.su-main-nav {
-  &.su-main-nav--light {
+.su-main-nav--light {
+  $_color-map: $main-nav-light-colors;
```
sherakama: Please namespace this variable.

core/src/scss/utilities/functions/_color.scss
```diff
@@ -0,0 +1,34 @@
...
+    @else {
+      @warn "Check that `#{$color}` is in color map $core-colors."
+    }
+  }
+  @else {
+    @if map-has-key($colormap, $color) {
+      $color: map-get($colormap, $color);
+    }
+    @else {
+      @warn "Check that `#{$color}` is in color map."
```
sherakama: Why not error out? I think we would want to have this break if a color is not available.

core/src/scss/utilities/functions/_color.scss
```diff
@@ -0,0 +1,34 @@
...
+// Style guide: Functions.Color
+//
+
+@function color($color, $colormap: $core-colors) {
+  @if $colormap == $core-colors {
+    @if map-has-key($core-colors, $color) {
+      $color: map-get($core-colors, $color);
+    }
+    @else {
+      @warn "Check that `#{$color}` is in color map $core-colors."
```
sherakama: Why not error out? I think we would want to have this break if a color is not available.

core/src/scss/utilities/variables/components/_main-nav.scss
```diff
@@ -0,0 +1,151 @@
...
+//    divider:                                $color-cool-grey,
+//    link--mobile:                           $color-white,
+//    link--mobile-hover:                     $color-white,
+//    border--mobile-hover:                   $color-white,
+//    link--mobile-active:                    $color-white,
+//    border--mobile-active:                  $color-vivid-red,
+//    link--mobile-expanded:                  $color-white,
+//    border--mobile-expanded:                transparent,
+//    border--mobile-active-expanded:         $color-white,
+//    light: (
```
sherakama: These need to be pulled out and reflected with the new multiple variable option.

core/src/scss/utilities/variables/components/_main-nav.scss
```diff
@@ -0,0 +1,151 @@
...
+
+  link--mobile-active:                    $color-white,
+  border--mobile-active:                  $color-vivid-red,
+
+  link--mobile-expanded:                  $color-white,
+  border--mobile-expanded:                transparent,
+  border--mobile-active-expanded:         $color-white,
+) !default;
+
+$main-nav-light-colors: (
```
sherakama: These should have their own KSS comment.

core/src/scss/utilities/variables/core/_core-color-maps.scss
```diff
@@ -0,0 +1,46 @@
+@charset "UTF-8";
+
+//
+// $core-colors
+//
+// SASS color map for the core Decanter theme. In the future, there might be more than one color theme.
```
sherakama: I think you can remove the part about themes.

core/src/scss/utilities/variables/components/_main-nav.scss
yvonnetangsu: Ooops...meant to do that. I'll fix this and make the other changes. Thanks!

core/src/scss/utilities/variables/components/_main-nav.scss
yvonnetangsu: Added separate section with markup/comments for each variants

core/src/scss/utilities/functions/_color.scss
yvonnetangsu: @sherakama , I changed it to @error but when I tried putting in a wrong color name, it doesn't show my custom error - any idea why?

core/src/scss/components/main-nav/_main-nav--light.scss
yvonnetangsu: Changed from $_color-map to $_su-main-nav-color-map (combining both @sherakama and @JBCSU 's suggestions)

core/src/scss/utilities/functions/_color.scss
sherakama: This may be something to do with the compiler and not the sass itself. The way we build in webpack may be hiding the error string.

sherakama:
Not sure why CC is complaining about the line length. I'm pretty sure we have that set to 120.

yvonnetangsu:
> Not sure why CC is complaining about the line length. I'm pretty sure we have that set to 120.

Yeah, CC's been weird lately. It complains about the braces not on their own lines but they are :\

PR #421, opened 2019-06-07
Update tags with HTML5 semantic tags.

rebeccahongsf:
# READY FOR REVIEW

# To Do:
- [ ] Update Components div tagd to HTML5 semantic tags. 
- [x] Update Card component.
- [x] Update Quote component.
- [x] Update Alert component - https://github.com/SU-SWS/decanter/pull/481
- [x] Update Brand Bar component.

# Summary
- Updating Components to contain HTML5 semantic tags.

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Compile styleguide
3. Review documentation in the components section (Section 3 Components) 

- Decanter

- @cjwest 
- https://stanfordits.atlassian.net/browse/THEME-15

# See Also


core/src/templates/components/card/card.twig
```diff
@@ -23,14 +23,14 @@
  * - card_button_label: The text label of the button component.
  */
 #}
-<div{{ attributes }} class="su-card {{ modifier_class }}">
+<article{{ attributes }} class="su-card {{ modifier_class }}">
```
sherakama: > The article element represents a component of a page that consists of a self-contained composition in a document, page, application, or site and that is intended to be independently distributable or reusable, e.g. in syndication. This could be a forum post, a magazine or newspaper article, a blog entry, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.

> W3C Specification: http://www.w3.org/TR/html5/semantics.html#the-article-element

I would agree then that the card should/can be an article.

sherakama: From http://html5doctor.com/the-article-element/

> The smell test for going independent
> An independent piece of content, one suitable for putting in an <article> element, is content that makes sense on its own.

core/src/templates/components/card/card.twig
```diff
@@ -64,7 +64,7 @@
 
   {# Big ol button. Adds a BEM class for targeting. #}
   {%- if card_button_label is not empty %}
-    <div class="su-card__button">
+    <button class="su-card__button">
```
sherakama: I don't think this makes sense as we are using a `link` template inside.

core/src/templates/components/quote/quote.twig
```diff
@@ -1,8 +1,8 @@
-<div class="su-quote">
+<article class="su-quote">
```
sherakama: This passes the smell test.

core/src/templates/components/card/card.twig
yvonnetangsu: I'm ok with the card being an article as well :)

core/src/templates/components/card/card.twig
yvonnetangsu: Yup, we can't change this one to `<button>` since there is a `<a>` nested inside (the link.twig). However, I wonder if we could get rid of the wrapper `<div class="su-card__button">` and just have the include with the link.twig? Perhaps the `div` was there for easier alignment?

core/src/templates/components/quote/quote.twig
yvonnetangsu: I'm ok with this one also!

JBCSU: I disagree. Quotes often depend on their context for proper comprehension. I don't believe a quote would "make sense on its own in a feed reader".

kgcreative: Why not `blockquote`?

yvonnetangsu: @kgcreative , I think we didn't use blockquote around the whole component itself because it has quite a bit of extra stuff - image, company/title and the name is used as a header (which Level Access asked us to change it to an actual h header). However, this makes me think that perhaps we should use blockquote for the actual quote portion of the component :)

JBCSU:
I disagree with changing the quote to use `<article>`. I don't think quotes are stand alone content - I think they almost always need to be taken in context.

core/src/templates/components/quote/quote.twig
JBCSU: I disagree with changing the quote to use <article>. I don't think quotes are stand alone content - I think they almost always need to be taken in context.

sherakama: I can buy that argument. Would `<section>` be more appropriate then?

yvonnetangsu: <section> sounds good to me. @JBCSU ?

yvonnetangsu: Another idea is to use `<figure>` - I know on stanford.edu, the event cards are figure elements. I looked up what it really means and it says [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure):
"The HTML <figure> (Figure With Optional Caption) element represents self-contained content, potentially with an optional caption, which is specified using the (<figcaption>) element. The figure, its caption, and its contents are referenced as a single unit." 
"Usually a `<figure>` is an image, illustration, diagram, code snippet, etc., that is referenced in the main flow of a document, but that can be moved to another part of the document or to an appendix without affecting the main flow."

core/src/templates/components/card/card.twig
sherakama: As this is the wrapper tag, we may have to wait until version 6 to get this out. This would break the 5.x API. I'll cut a new v6 branch and point this PR to it.

core/src/templates/components/quote/quote.twig
yvonnetangsu: @cjwest Actually, the more I think about this, `<figure>` actually seems like a good choice here. I'm working on the WordPress template, and their Gutenberg WYSIWYG editor also output a `<figure>` element for their fancy quote component. They have a `<blockquote>` (for the quote and cite portion) inside a `<figure>`.

cjwest: @yvonnetangsu,
FYI, I am excited we are exploring this topic!
So you're suggesting that for the quote we use `<figure>`?
``` 
<figure class="su-quote">
```
I'm not sure it's fitting semantically. Also, I don't think that an image should be required for a quote so having a figure without a image would be weird. The body or quote, however, should be required.
 I'd welcome other opinions on this.

yvonnetangsu: @cjwest , apparently `<figure>` isn't only for elements with an image, for example, a lot of article comments are individually wrapped within a `<figure>` tag also. WordPress' quote block and video blocks also use `<figure>` and there are no images involved.

yvonnetangsu: Check out Smashing Magazine's individual comments for example:
https://www.smashingmagazine.com/2018/06/bem-for-beginners/

I'm glad we're having a discussion on this as well! :)

cjwest: Here's an example where they combine a figure with a block quote.

https://github.com/hail2u/html-best-practices#dont-include-attribution-directly-in-blockquote-element

What are your thoughts on this:
```
<figure>
  <img class="su-quote__img" src="{{ src }}" alt="{{ alt }}">
   <p class="su-quote__heading">{{ name }}</p>
  <blockquote>
    <p class="su-quote__quote">"{{ text }}"</p>
  </blockquote>
  <figcaptionclass="su-quote__bio">{{ bio }}</figcaption>
</figure>
```

yvonnetangsu: I need a little more time to think about this but I'm going on vacation tomorrow 😄 . I know I don't like the one with the nested `figure` though - don't think that's allowed.

cjwest: Actually, you can nest '<figure>`

core/src/templates/components/quote/quote.twig
```diff
@@ -1,8 +1,10 @@
...
+    <h3 class="su-quote__heading">{{ name }}</h3>
     <p class="su-quote__bio">{{ bio }}</p>
-    <p class="su-quote__quote">"{{ text }}"</p>
-  </div>
-</div>
+    <blockquote>
+      <p class="su-quote__quote">"{{ text }}"</p>
+    </blockquote>
+  </section>
+</figure>
```
cjwest: Okay @sherakama, @yvonnetangsu, @JBCSU,
Here's a proposed HTMP5 approach for the on the quote. Your thoughts?

sherakama: I don't think `figure` makes sense here.  

From: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure
> Usually a <figure> is an image, illustration, diagram, code snippet, etc., that is referenced in the main flow of a document, but that can be moved to another part of the document or to an appendix without affecting the main flow.

The quote is not an image, illustration, or diagram to me.

sherakama: Ok, I just read through your comment thread. I still have mixed feelings about the figure tag but there seems to be a strong case for it. Ok.

sherakama:
Please add to the CHANGELOG.txt this change and then this is GTG.

cjwest:
@sherakama,
I believe this is ready to merge.

sherakama:
Thank you!.

Issue #381, opened 2019-04-10
Review flex grid classes for DX.


buttonwillowsix:

buttonwillowsix:
Can we keep responsive gutters?
Can we do it? Do a quick prototype and implement if possible.

If it is in conflict, the developer experience is priority (i.e. not pass 6 classes to keep responsive gutters.)

sherakama:
Another example of a Styleguide/Pattern library with a bespoke grid system similar to Bootstrap: http://walmartlabs.github.io/web-style-guide/

sherakama:
Example of a grid system with responsive gutters: 
https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-margins

PR #425, opened 2019-06-10
DEC-381: Responsive flex classes with responsive gutters

sherakama:
# READY FOR REVIEW

# Summary
- Improves the developer experience in using the flex based classes so that the developer does not have to add the size for every breakpoint
- Breakpoint based flex size now has responsive behavior for all sizes above the one in use.

# Needed By (Date)
- End of the week

# Urgency
- Medium

# Steps to Test

1. Review code
2. Compile sass
3. Compile styleguide
4. Review core flex-box example for consistency with the existing version
5. Review flex-box layouts for issues

Test new behavior:
1. Create a new flexbox container and child element
2. Add different flexbox classes and review for expected behavior

- Flex classes
- Flex layouts

- #381 

# See Also


core/src/templates/layout/flex/three-column/battleship/battleship.twig
```diff
@@ -24,49 +24,49 @@
...
     {% if first is not empty %}
-    <div class="flex-lg-8-of-12 flex-xl-8-of-12 flex-2xl-8-of-12">
+    <div class="flex-lg-8-of-12">
       {{ first }}
     </div>
     {% endif %}
 
     {% if second is not empty %}
-    <div class="flex-md-6-of-12 flex-lg-4-of-12 flex-xl-4-of-12 flex-2xl-4-of-12">
+    <div class="flex-md-6-of-12 flex-lg-4-of-12">
```
yvonnetangsu: Haven't taken a really close look but Battleship has some issues - something to do with it's md-6 but lg-4?

sherakama: Oh, this might be because the 4-of-12 is rendered before the 6-of-12 items as the loop goes through grid first then breakpoint. I'll refactor to see if that helps.

core/src/templates/layout/flex/four-column/molive/molive.twig
```diff
@@ -30,19 +30,19 @@
 
     {# Because of the responsive behaviour each column is required. #}
     <section class="flex-container centered-container">
-      <div class="flex-md-6-of-12 flex-lg-6-of-12 flex-xl-3-of-12 flex-2xl-3-of-12">
+      <div class="flex-md-6-of-12 flex-xl-3-of-12">
         {{ first }}
       </div>
 
-      <div class="flex-md-6-of-12 flex-lg-6-of-12 flex-xl-3-of-12 flex-2xl-3-of-12">
+      <div class="flex-md-6-of-12flex-xl-3-of-12">
```
yvonnetangsu:
```suggestion
      <div class="flex-md-6-of-12 flex-xl-3-of-12">
```

sherakama:
@yvonnetangsu ready for another pass.

yvonnetangsu:
Found one thing about flex grid that's on the live site also, so it's not because of your work in this branch. Under Core - > flex grid, the cells that say x-of-12-for-all seem to behave the same as x-of-12-for-sm-and-up, ie, they don't really do x-of-12 at xs. Is this intentional?

Screen Shot 2019-06-11 at 11 23 11 AM

yvonnetangsu:
For the push classes, looks like .flex-push-xs-1, .flex-push-sm-1, .flex-push-md-1 etc all do the same thing (set the same margin-left percentage). Could we just have .flex-push-1 etc (get rid of the breakpoint specific classes)?
Screen Shot 2019-06-11 at 11 53 15 AM

sherakama:
> I can see that this makes sense (at XS, everything takes up 100% and just stack on top of each other). In that case, perhaps we could eliminate the .flex-sm-x-of-12 classes since they're the same as the .flex-x-of-12 ones?

Perhaps we can now, prior to this branch, if you had a `flex-6-of-12` and a `flex-sm-3-of-12` you would get 3-of-12 only on the small and 6-of-12 everywhere else. Now, with the growing cascade, they pretty much do the same.

sherakama:
> For the push classes, looks like .flex-push-xs-1, .flex-push-sm-1, .flex-push-md-1 etc all do the same thing (set the same margin-left percentage). Could we just have .flex-push-1 etc (get rid of the breakpoint specific classes)?

No, I don't think so, as they activate at different breakpoints. eg: push at md but not at lg.

yvonnetangsu:
> > For the push classes, looks like .flex-push-xs-1, .flex-push-sm-1, .flex-push-md-1 etc all do the same thing (set the same margin-left percentage). Could we just have .flex-push-1 etc (get rid of the breakpoint specific classes)?
> 
> No, I don't think so, as they activate at different breakpoints. eg: push at md but not at lg.

Good point!

sherakama:
@yvonnetangsu I removed the `flex-sm-x-of-y` classes and refactored the collapse and gap code to not assume our breakpoint codes as in the future we might want to add an `xxs` or something.

yvonnetangsu:
> @yvonnetangsu I removed the `flex-xs-x-of-y` classes and refactored the collapse and gap code to not assume our breakpoint codes as in the future we might want to add an `xxs` or something.

Thanks for the change, @sherakama ! This looks good. I do think I need a bit more time to test this out more thoroughly. I'll look at it some more tonight.

core/dist/css/decanter-grid.css
```diff
@@ -674,498 +674,1590 @@ body {
...
+  .flex-sm-11-of-12 {
     -webkit-box-flex: 0;
         -ms-flex: 0 0 calc(91.6666666667% - 4px);
             flex: 0 0 calc(91.6666666667% - 4px);
     max-width: calc(91.6666666667% - 4px); } }
 
-@media only screen and (min-width: 0) and (max-width: 575px) {
-  .flex-xs-12-of-12 {
+@media only screen and (min-width: 576px) {
+  .flex-sm-12-of-12 {
```
yvonnetangsu: Is it possible to do this optimization - for all the `.flex-xx-12-of-12` classes, only write the first media query `@media only screen and (min-width: 576px)` since all the subsequent media queries generate the same numbers?
Screen Shot 2019-06-12 at 1 55 17 AM

core/dist/css/decanter-grid.css
```diff
@@ -674,498 +674,1590 @@ body {
...
+    max-width: calc(8.3333333333% - 36.6666666667px); } }
+
+@media only screen and (min-width: 1500px) {
+  .flex-2xl-1-of-12 {
+    -webkit-box-flex: 0;
+        -ms-flex: 0 0 calc(8.3333333333% - 44px);
+            flex: 0 0 calc(8.3333333333% - 44px);
+    max-width: calc(8.3333333333% - 44px); } }
+
+@media only screen and (min-width: 576px) {
```
yvonnetangsu: It looks like for example, the `.flex-xl-x-of-12` classes start with media query for the `xl` breakpoint (which is right), but the `.flex-2xl-x-of-12` classes start with media query for the `xs` breakpoint? Seems like they should just start at `2xl` breakpoint?
Screen Shot 2019-06-12 at 1 59 49 AM

yvonnetangsu: Yup, something's weird with the `2xl `and up classes 👀  Screenshot is at `sm` breakpoint.
Screen Shot 2019-06-12 at 2 14 53 AM

yvonnetangsu:
Got a question about the push classes, @sherakama. Posting here since you're concentrating 😃 
Am I doing something wrong here? It looks like the push value is a bit off? Also, do we need the xs push classes since everything is 100% width?

Screen Shot 2019-06-12 at 10 18 51 AM

sherakama:
> Got a question about the push classes, @sherakama. Posting here since you're concentrating 😃
> Am I doing something wrong here? It looks like the push value is a bit off? Also, do we need the xs push classes since everything is 100% width?

Good questions. I will review.

core/dist/css/decanter-grid.css
yvonnetangsu: This one works great - thanks!

Issue #423, opened 2019-06-10
Investigate the best way to include custom icons into our system


kerri-augenstein:
Look at various ways to enable using icons. 

**Requirements**
- Easy to change the color of the icon using css 
- Use custom icons created by designers in Figma

yvonnetangsu:
We're going with the `mask` method so we can change the svg icon colors using `background-color` in css. However, IE doesn't understand `mask`, so we'll be hiding the icons in IE since they are not absolutely necessary.

PR #426, opened 2019-06-10
Use svg icons as "mask" so their colors could be changed using CSS background-color

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Investigate best way to use SVG icons
https://github.com/SU-SWS/decanter/issues/423
- Use SVG icons as mask seems to do the trick (except IE). After discussion with Kerri and JB, we think it's ok to just hide the icon in IE.
- This is just a test so I've only done work on .su-link--external. If this way of including SVG icon is approved, we'll refactor the other components where SVG icons are used in another PR.

# Needed By (Date)
- Sooner the better since Kerri/JB is working on the new link components this week.

# Urgency
- Earlier the better :)

# Steps to Test

2. Go to Components - links, and check that the external link variant icon works fine in all browsers except IE, and that the icon is hidden in IE.

- Decanter

- https://github.com/SU-SWS/decanter/issues/423


core/src/scss/utilities/mixins/link/_external-link.scss
```diff
@@ -13,17 +13,22 @@
...
   }
 
   &:hover::after {
-    background-image: url("#{$image-path}/#{$external-link-hover}.png");
-    background-image: url("#{$image-path}/#{$external-link-hover}.svg");
+    background-color: $color-black;
+  }
+
+  // IE only - hides icon.
+  @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
sherakama: Can we make this a mixin? This seems like something we would want to standardize and have available.

yvonnetangsu: Sounds like a great idea :)

core/src/scss/utilities/mixins/link/_external-link.scss
```diff
@@ -13,17 +13,22 @@
...
-    background: url("#{$image-path}/#{$external-link}.png") no-repeat 0 0;
-    background: url("#{$image-path}/#{$external-link}.svg") no-repeat 0 0;
-    background-size: 100%;
-    content: '';
+    @include size(0.65em);
     display: inline-block;
-    height: 0.65em;
-    width: 0.65em;
+    content: '';
+    mask: url("#{$image-path}/#{$external-link}.svg") no-repeat 0 0;
```
sherakama: Browser support for this looks limited: https://caniuse.com/#feat=css-masks
Did you check that this spec meets our compatibility requirements? https://github.com/SU-SWS/decanter/wiki/Compatibility

core/src/scss/utilities/mixins/link/_external-link.scss
```diff
@@ -13,17 +13,22 @@
...
+    mask-size: cover;
```
sherakama: This doesn't look well supported: https://developer.mozilla.org/en-US/docs/Web/CSS/mask-size#Browser_compatibility

yvonnetangsu:
> Can you verify that the browser support requirements for Decanter are met?

Yeah, I've tested on desktop Firefox, Safari, Chrome, Edge - those are all ok. I'll test it on Android and iOS.

yvonnetangsu:
> > Can you verify that the browser support requirements for Decanter are met?
> 
> Yeah, I've tested on desktop Firefox, Safari, Chrome, Edge - those are all ok. I'll test it on Android and iOS.

@sherakama , I checked a few devices on Browserstack picking not the newest phones on purpose - Galaxy S8 v.7 Chrome, iOS 11 iPhone SE Safari, iOS 10 iPhone 7 Chrome, Pixel 2 Android 8 Firefox and all ok.

yvonnetangsu:
> Thank you for testing. This is GTG then.

Cool - I'm writing that IE-only mixin. Should be fairly quick.

JBCSU:
Our compatibility guidelines say we will support the latest version of Edge and 1 back. But the mask only seems to work on the latest version of Edge, i.e. 18. It doesn't render on Edge 17 or 16. I'm totally fine with that - just sayin'. Otherwise, (almost) all the device / browser combinations I tested behave as expected. I had trouble with the default (Samsung) browser on a Galaxy S8, but it was not related to this change.

yvonnetangsu:
> Our compatibility guidelines say we will support the latest version of Edge and 1 back. But the mask only seems to work on the latest version of Edge, i.e. 18. It doesn't render on Edge 17 or 16. I'm totally fine with that - just sayin'. Otherwise, (almost) all the device / browser combinations I tested behave as expected. I had trouble with the default (Samsung) browser on a Galaxy S8, but it was not related to this change.

Yeah, I found out about that last night. Though https://caniuse.com/#search=mask says it can be enabled in Edge 17 behind the flag? Any idea what that means? Is that a setting in the browser?

I'm still hoping that this is ok, since Edge is going Chromium soon and I like the ease of being able to change the colors easily. What do you think? 🤔

yvonnetangsu:
> Our compatibility guidelines say we will support the latest version of Edge and 1 back. But the mask only seems to work on the latest version of Edge, i.e. 18. It doesn't render on Edge 17 or 16. I'm totally fine with that - just sayin'. Otherwise, (almost) all the device / browser combinations I tested behave as expected. I had trouble with the default (Samsung) browser on a Galaxy S8, but it was not related to this change.

Perhaps we do this in phases. We use this approach for the more decorative icons (e.g., external link), and for components like Main Nav (which it's more important to have the icons visible), we switch to this approach after a newer version of Edge gets pushed out and IE population drops even lower ?

JBCSU:
> https://caniuse.com/#search=mask

JBCSU:
> [...] https://caniuse.com/#search=mask says it can be enabled in Edge 17 behind the flag? Any idea what that means? Is that a setting in the browser?

[MS Edge 17 doc says](https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/whats-new/edgehtml-17):

> EdgeHTML 17 includes experimental support for CSS Masking. The partial implementation introduces the CSS mask-image and mask-size properties. Check the "Enable CSS Masking" flag in about:flags to being experimenting!

No Edge 17 users will do this, so it's effectively unsupported.

> I'm still hoping that this is ok, since Edge is going Chromium soon and I like the ease of being able to change the colors easily. What do you think? 🤔

I'm fine with it being only the latest version of Edge. @sherakama, how about you?

yvonnetangsu:
Hey @JBCSU did you close this PR? 👀

core/src/scss/utilities/variables/core/_core-color-maps.scss
```diff
@@ -37,10 +37,10 @@ $core-colors: (
   text--high-contrast:          #000,
   text--reverse:                $color-white,
 
+  link:                         $color-bright-blue,
+  link--hover:                  $color-black,
```
JBCSU: I chose to replace `link` and `link--hover` with `link-inline` and `link-inline--hover` to be more explicit about the application. I don't want to have both, so let's decide. I prefer the `-inline` variants, in which case please use those. But if you feel strongly that `link` and `link--hover` are better, then please delete the ones I added and change all references to the `-inline` map entries to match

yvonnetangsu: I like having a clean default `link` color without any modifiers since we have clean `bg`, `text` colors. Also, the `-inline` makes me wonder if it only applies to `display:inline` links. For example, the links in the vertical link list will probably take on this default link color, but it might be a bit strange if I apply the color `link-inline` even though the link might be a block element.

JBCSU: Ok, then please delete the `-inline` options and update all references to them to refer to the shorter names.

yvonnetangsu: Done! Added a link--alt and link--alt-hover for cases like the secondary button link style. I feel like the bright red link will appear on some sites in some other ways so I make that the link--alt color.

JBCSU: Ok. I introduced an error in the doc for `@function color()`, so I fixed that.

yvonnetangsu: Thanks!!

yvonnetangsu:
> I think this is GTG.

Yay, merging before you changed your mind 😂

Issue #424, opened 2019-06-10
Change "Default link" to "Inline link", change colors


JBCSU:
The link style currently referred to **Default styling** actually refers to links that are inline with paragraph text. Change the name to **Inline link** to make this more obvious.

Also, the current inline link styling is red, normal font weight, and no underline (color only). This makes links very difficult to see, and it does not match Figma. Change the inline link style to `#006CB8`, 600 weight, underline. Hover state is same, but text color (`#2E2D29`).

Issue #341, opened 2019-03-13
Create additional link components


josephgknox:
- External link
- Download link 
- Video link
- Jump link
- Internal link

Source: https://www.figma.com/file/Kmd4utmJFPRMVeCFEEBQhLtx/Stanford-Design-Library?node-id=76%3A4256

<img width="184" alt="additional_link_styles" src="https://user-images.githubusercontent.com/26723613/59152402-42b8df80-8a11-11e9-8875-e0ff53c147fe.png">

Icons are documented in Figma here: 
https://www.figma.com/file/Kmd4utmJFPRMVeCFEEBQhLtx/Stanford-Design-Library?node-id=2213%3A544

sherakama:
Is this ticket a create more link components as additional components or is it a more link component as in read more?

josephgknox:
Thanks for asking for clarification here. I can see how this can read in a few ways. Ha! This is to add additional link components. I will rename the ticket, using the word "additional" instead of "more."

buttonwillowsix:
Have Kerri do design & Scope BEFORE decanterpalooza. Otherwise have Kerri spend time during decanterpalooza?

buttonwillowsix:
- make a more generic icon + link (with a few choices?)
- look at states for all links? (more link, etc.)

buttonwillowsix:
Dena to split into one ticket for looking at states and another ticket for looking at new link types.

yvonnetangsu:
Done

PR #427, opened 2019-06-10
424 default link style

JBCSU:
# READY FOR REVIEW

# Summary
Update the default styling for links. See #424.

# Needed By (Date)
Other link work in Decanterpalooza will build on this, so sooner is better than later.

# Urgency
Other link work in Decanterpalooza will build on this

# Steps to Test

1. Pull the branch
1. Rebuild the assets and the styleguide
1. Review the link styles.

Note that most of the diffs in this PR are because it's based on the `386-nav-color-map` branch. I've changed the following files

- `scss/components/alert/_alert.scss`
- `scss/utilities/functions/color.scss`
- `scss/utilities/mixins/button/_button-secondary.scss`
- `scss/utilities/mixins/link/_link.scss`
- `scss/utilities/placeholders/components/_cta.scss`
- `scss/utilities/variables/core/_core-color-maps.scss`
- `templates/components/link/link.json`
- `kss/builder/decanter/scss/_base_elements.scss`

- #341 will depend on this

- This PR depends on PR #420, as it uses the color maps defined in that PR. That PR should be approved and merged first.
- @kerri-augenstein @yvonnetangsu

# See Also


yvonnetangsu:
@JBCSU, looks like the font-weight is changed to semi-bold in several components, e.g., lockups, global footer, and the button variants of the links - John Holleman said font weight of buttons should be regular 400.

kerri-augenstein:
@JBCSU @yvonnetangsu given that we've effected so many other components (which I imagine may have been one of John's reasons for going with the Regular and non-underlined style to begin with), should we revert back, and add in a "default inline link" style, and let the default "body" style be as previously designed. 

I like this because... 

It does not force us to undo the Semi-bold and Underline across infinite numbers of links on our sites, which we would inevitably be forced to do. Lots of lines of code we could avoid. 

A nice simple default "body" style seems good to me. Regular weight and no underline.

yvonnetangsu:
> @JBCSU @yvonnetangsu given that we've effected so many other components (which I imagine may have been one of John's reasons for going with the Regular and non-underlined style to begin with), should we revert back, and add in a "default inline link" style, and let the default "body" style be as previously designed.
> 
> I like this because...
> 
> It does not force us to undo the Semi-bold and Underline across infinite numbers of links on our sites, which we would inevitably be forced to do. Lots of lines of code we could avoid.
> 
> A nice simple default "body" style seems good to me. Regular weight and no underline.

I agree with @kerri-augenstein. Could we revert the default link font-weight and have another variant for link within paragraph or something? :)

JBCSU:
> > @JBCSU @yvonnetangsu given that we've effected so many other components (which I imagine may have been one of John's reasons for going with the Regular and non-underlined style to begin with), should we revert back, and add in a "default inline link" style, and let the default "body" style be as previously designed.
> > I like this because...
> > It does not force us to undo the Semi-bold and Underline across infinite numbers of links on our sites, which we would inevitably be forced to do. Lots of lines of code we could avoid.
> > A nice simple default "body" style seems good to me. Regular weight and no underline.
> 
> I agree with @kerri-augenstein. Could we revert the default link font-weight and have another variant for link within paragraph or something? :)

1. I'm not sure how we would apply a "default inline link" style variant. I certainly don't think we want to require users to add a special class to every link they put anywhere in there content. But if they apply the class to a larger container, e.g `<body>`, `<main>`, `<article>`, then our other components would end up inheriting the styling anyway.
1.  More generally, our components are opinionated about how they look - we have a particular look we want to enforce. I think it's appropriate to be explicit about the styling of links in our components. Suppose a user decides they want their links to be bold even if Decanter's default link style doesn't include the font-weight. Then all our components would inherit that anyway.

We only have 14 components, and most of them work ok already. It's only a few components that need special attention. I'll just go ahead and be explicit about the link styling in those components.

yvonnetangsu:
> > > @JBCSU @yvonnetangsu given that we've effected so many other components (which I imagine may have been one of John's reasons for going with the Regular and non-underlined style to begin with), should we revert back, and add in a "default inline link" style, and let the default "body" style be as previously designed.
> > > I like this because...
> > > It does not force us to undo the Semi-bold and Underline across infinite numbers of links on our sites, which we would inevitably be forced to do. Lots of lines of code we could avoid.
> > > A nice simple default "body" style seems good to me. Regular weight and no underline.
> > 
> > 
> > I agree with @kerri-augenstein. Could we revert the default link font-weight and have another variant for link within paragraph or something? :)
> 
> 1. I'm not sure how we would apply a "default inline link" style variant. I certainly don't think we want to require users to add a special class to every link they put anywhere in there content. But if they apply the class to a larger container, e.g `<body>`, `<main>`, `<article>`, then our other components would end up inheriting the styling anyway.
> 2. More generally, our components are opinionated about how they look - we have a particular look we want to enforce. I think it's appropriate to be explicit about the styling of links in our components. Suppose a user decides they want their links to be bold even if Decanter's default link style doesn't include the font-weight. Then all our components would inherit that anyway.
> 
> We only have 14 components, and most of them work ok already. It's only a few components that need special attention. I'll just go ahead and be explicit about the link styling in those components.

That's a good point, @JBCSU

JBCSU:
Ok, I think I've fixed all the cascades. Can someone with a keener eye for fonts than I double check, please? @yvonnetangsu

yvonnetangsu:
> Ok, I think I've fixed all the cascades. Can someone with a keener eye for fonts than I double check, please? @yvonnetangsu

Font weight looks good, @JBCSU , but what happens to the underline? I thought we're adding underline to the default link style but I could be wrong 🤔 . Also, I believe the links in alerts have underlines to begin with.

yvonnetangsu:
One more thing, the :focus state of the default link needs to be changed to black color also (should match the :hover state when keyboard user tab to it).

JBCSU:
> > Ok, I think I've fixed all the cascades. Can someone with a keener eye for fonts than I double check, please? @yvonnetangsu
> 
> Font weight looks good, @JBCSU , but what happens to the underline? I thought we're adding underline to the default link style but I could be wrong 🤔 . Also, I believe the links in alerts have underlines to begin with.

The new default link style does have underline. I therefore removed the specification from links inline in alert text and allow those links to inherit the default style, as I believe those links should actually get the default inline link style (i.e. if a developer overrides the default inline styles).

JBCSU:
> One more thing, the :focus state of the default link needs to be changed to black color also (should match the :hover state when keyboard user tab to it). And the animation - those apply to the :focus as well.

`:focus` was not in the css before, but I can add it.

Animation is not part of the default inline link style, which is this PR. If we need to address animation, I propose we do it in #341.

yvonnetangsu:
> > > Ok, I think I've fixed all the cascades. Can someone with a keener eye for fonts than I double check, please? @yvonnetangsu
> > 
> > 
> > Font weight looks good, @JBCSU , but what happens to the underline? I thought we're adding underline to the default link style but I could be wrong 🤔 . Also, I believe the links in alerts have underlines to begin with.
> 
> The new default link style does have underline. I therefore removed the specification from links inline in alert text and allow those links to inherit the default style, as I believe those links should actually get the default inline link style (i.e. if a developer overrides the default inline styles).

Great - yeah, turns out I had to rebuild the styleguide. Looks good to me.

yvonnetangsu:
> > One more thing, the :focus state of the default link needs to be changed to black color also (should match the :hover state when keyboard user tab to it). And the animation - those apply to the :focus as well.
> 
> `:focus` was not in the css before, but I can add it.
> 
> Animation is not part of the default inline link style, which is this PR. If we need to address animation, I propose we do it in #341.

Interesting....I didn't realize the current live site doesn't have focus styles for the links. I must be thinking about the homesite (the jump links have animation when you tab to them). We could merge this in, and I can deal with the focus state in the animation branch later?

yvonnetangsu:
One probably final thing, JB - the "Decanter" logo now has underline and the font weight is different. Could you please remove the underline and change it to font-weight: 700?

JBCSU:
I did go ahead and `:focus` states, and fixed all the cascades' `:focus` states. And I fixed the link in the masthead. Anything else, @yvonnetangsu? Thanks for your very careful eye!

kss/builder/decanter/kss-assets/css/kss.css
```diff
@@ -1684,7 +1684,11 @@ body {
     -ms-grid-row: 1;
     -ms-grid-column: 1; }
     .masthead .brand-logo a {
-      color: #ffffff; }
+      color: #ffffff;
+      text-decoration: none;
+      font-weight: 700; }
+      .masthead .brand-logo a:hover, .masthead .brand-logo a:focus {
```
yvonnetangsu: Do we want a hover/focus color for the Decanter logo? This might be a @kerri-augenstein question?

JBCSU: I did put a hover / focus color on it. I used `$color-cloud`, which I had found in the alert variants that use white text on a dark(ish) background. Not a great difference, IMO, but the precedent was there.

yvonnetangsu: Yeah, the color is ok, but since we don't have hover state for the lockup or sitename (Redwood), perhaps we just don't do color change for these logo type things for consistency?

yvonnetangsu:
GTG for me! The Decanter logo hover color - I can live with either.

JBCSU:
Great, thanks @yvonnetangsu & @sherakama. Merging now.

PR #439, opened 2019-06-11
Create Masthead component

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Create the masthead component

# Needed By (Date)
- When it's ready

# Urgency
- N/A

# Steps to Test

2. Look at the masthead variants at all breakpoints

- Decanter
- Redwood WordPress

- #439


yvonnetangsu:
Also requesting your review in advance, @sherakama . I still need to add all the variables properly from the nested components and some fixup I need to do, but please leave any comments. I'm thinking for the different brand bar colors the developers can just pass in the brand bar modifier class themselves instead of having those as masthead variants.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+    @include centered-column;
+    @include modular-spacing('padding', 1 null null);
+  }
+
+  .su-lockup {
+    @include modular-spacing('margin', null null 0);
+    max-width: fit-content;
+
+    @include grid-media-max('md') {
+      @include flex-column($columns: 10);
```
sherakama: No default allowed in calling a mixin. 

`@include flex-column($columns:);`
or
`@include flex-column(10);`

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+      // On mobile, lockup needs a higher z-index than nav because the nav takes up 100% width including margin
+      z-index: 10010;
+    }
+
+    @include grid-media('lg') {
+      @include flex-column($columns: 8);
```
sherakama: You cannot pass default values to a mixin call.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  .su-lockup {
+    @include modular-spacing('margin', null null 0);
+    max-width: fit-content;
+
+    @include grid-media-max('md') {
+      @include flex-column($columns: 10);
+      // On mobile, lockup needs a higher z-index than nav because the nav takes up 100% width including margin
+      z-index: 10010;
```
sherakama: @buttonwillowsix z-index here.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  .su-site-search {
+    @include grid-media-max('md') {
+      display: none;
+    }
+
+    @include grid-media('lg') {
+      @include flex-column($columns: 4);
```
sherakama: You cannot pass default values to a mixin call.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+
+  @include grid-media-max('sm') {
+    .su-main-nav {
+      @include margin(-57px null null);
+    }
+  }
+
+  @include grid-media-only('md') {
+    .su-main-nav {
+      @include flex-column($columns: 2);
```
sherakama: You cannot pass default values to a mixin call.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+// .su-masthead--light - Masthead with the light variant of main nav
+// .su-masthead--dark - Masthead with a dark color theme
+//
+// Markup: ../templates/components/masthead/masthead.twig
+//
+// Style guide: Components.Masthead
+//
+
+.su-masthead {
+  @include box-shadow($depth: 'shallow', $stroke: 0);
```
sherakama: You cannot pass default values to a mixin call.

core/src/scss/components/masthead/_masthead.scss
```diff
@@ -0,0 +1,125 @@
...
+    .su-main-nav {
+      @include flex-column($columns: 2);
+      @include margin(auto null null);
+    }
+  }
+
+  .su-main-nav {
+    @include grid-media-max('md') {
+      > ul {
+        @include box-shadow($depth: 'medium', $stroke: none);
```
sherakama: You cannot pass default values to a mixin call.

core/src/scss/utilities/mixins/shadow/_box-shadow.scss
```diff
@@ -11,6 +11,9 @@
 // Style guide: Mixins.Shadow.Box Shadow
 //
 @mixin box-shadow($depth: 'flat', $stroke: '1px solid #C6C6C6') {
-  border: #{$stroke};
+  @if $stroke != 0 and $stroke != 'none' and $stroke != null {
```
sherakama: Why the extra validation?

yvonnetangsu: > Why the extra validation?

Oh, that's because sometimes I just want the shadow and no border, and if I use the mixin it will write `border: 0`. Sometimes that overrides other custom borders styles I have on those elements on other sites, so I'd rather it doesn't output a border property at all when there's no border.

core/src/scss/components/masthead/_masthead.scss
yvonnetangsu: Good to know. All fixed! 😄

Issue #437, opened 2019-06-11
Card variant for “non-card” style - no background container (vertical + horizontal)


buttonwillowsix:
remove border
remove background color
remove shadow 

For more context: 
https://docs.google.com/document/d/1BVBr71fxqSEzxP-yb3MAQ_5Dm2i9bY9KydJZfV_Aaxk/edit#

yvonnetangsu:
Are we calling this variant .su-card--fake? 😄

kerri-augenstein:
@JBCSU think we should screenshare after our standup to review, as I'm not setup to see the changes.

Issue #436, opened 2019-06-11
Add background color to card component (vertical + horizontal)


buttonwillowsix:
Existing cards do not include a background color, and they should

Issue #435, opened 2019-06-11
Icon variants for vertical card + horizontal card


buttonwillowsix:
Make changing the size of media possible (scale up or down the icons, without dead space) 

For more context on this issue: https://docs.google.com/document/d/1BVBr71fxqSEzxP-yb3MAQ_5Dm2i9bY9KydJZfV_Aaxk/edit#

PR #440, opened 2019-06-12
436 437 cards

JBCSU:
# READY FOR REVIEW

# Summary
- #436 Add background color to card contents
- #437 Add `.su-card--minimal` variant

# Needed By (Date)
- End of week

# Urgency
- Useful for #435 icon variants for cards

# Steps to Test

1. Pull the branch
1. Rebuild assets
1. Rebuild style guide
1. Check out the new `.su-card--minimal` variant of a card
1. Use the inspector to add `background-color: cornsilk;` to the `.kss-modifier__wrapper` that contains all the card variants.
    - Note that vertical and horizontal cards have a white background for the card contents.
    - Note that `.su-card--minimal` does not have a background color.

- Useful for #435 icon variants for cards

- #435 icon variants for cards

# See Also


yvonnetangsu:
.su-card--minimal is a great name, btw 😄 . Better than .su-card--fake 😸

JBCSU:
@yvonnetangsu I've removed the appropriate padding from the default (vertical) and horizontal cards.

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,20 @@
+.su-card--minimal {
+  @include box-shadow($depth: 'flat', $stroke: 'none');
+  background-color: inherit;
+
+  .su-card__contents {
+    // override modular spacing at all breakpoints
+    @include grid-media-only(all) {
+      padding-left: 0;
+      padding-right: 0;
```
yvonnetangsu: For the vertical card, I understand keeping the padding-top so there's space between the image and the content, but we probably should set `padding-bottom: 0` also, since we don't add any padding/margin on the "outside" of any components in Decanter.

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,20 @@
...
+    }
+  }
+
+  &.su-card--horizontal {
+    // override modular spacing at all breakpoints
+    @include grid-media-only(all) {
+      padding-left: 0;
+      padding-right: 0;
+    }
```
yvonnetangsu: I think we could just do `padding: 0` for all sides for the horizontal minimal variant.

core/src/scss/components/card/_card--minimal.scss
JBCSU: done

core/src/scss/components/card/_card--minimal.scss
JBCSU: done

core/src/scss/components/card/_card--minimal.scss
yvonnetangsu: Actually could we use the padding mixin here
`@include padding(null 0 0);`
Sorry I forgot about that

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,20 @@
...
+      padding-left: 0;
+      padding-bottom: 0;
+      padding-right: 0;
+    }
+  }
+
+  &.su-card--horizontal {
+    // override modular spacing at all breakpoints
+    @include grid-media-only(all) {
+      padding: 0;
```
yvonnetangsu: And same here
`@include padding(0);`
I believe we're using these mixins for consistency for everything in Decanter.

yvonnetangsu:
Actually I should be clearer - Kerri have the button left-aligned if the headlines and text are left-aligned, the button is center-aligned if the text is center-aligned.

core/src/scss/components/card/_card--minimal.scss
JBCSU: I actually prefer just specifying the padding directly. IMO the mixin saves very little typing and makes it slightly harder to read.

yvonnetangsu: > I actually prefer just specifying the padding directly. IMO the mixin saves very little typing and makes it slightly harder to read.

That might be a question for @sherakama ?  I don't particularly mind either way myself, but I believe at one point we agree that we'll use these mixins (padding and margin) wherever we can in Decanter 🤔

JBCSU: l changed it to use the mixin, and I left-aligned the button.

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,22 @@
+.su-card--minimal {
```
sherakama:
```suggestion
@charset 'UTF-8';

.su-card--minimal {
```

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,22 @@
...
+  @include box-shadow($depth: 'flat', $stroke: 'none');
+  background-color: inherit;
+
+  .su-card__contents {
+    // override modular spacing at all breakpoints
```
sherakama:
```suggestion
    // Override modular spacing at all breakpoints
```

Please to forgive my pedantry.

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,22 @@
...
+      @include padding(null 0 0);
+    }
+  }
+
+  .su-card__button {
+    text-align: left;
+  }
+
+  &.su-card--horizontal {
+    // override modular spacing at all breakpoints
```
sherakama:
```suggestion
    // Override modular spacing at all breakpoints
```
Please to forgive my pedantry.

core/src/scss/utilities/mixins/breakpoint/_grid-media-only.scss
```diff
@@ -1,16 +1,25 @@
...
 //
-// $breakpoint - The breakpoint shortcode. Eg: xs
+// $breakpoint - The breakpoint key code eg: 'sm', or 'all' for all breakpoints.
 //
 // Style guide: Mixins.Breakpoint.grid-media-only
 //
 @mixin grid-media-only($breakpoint: 'xs') {
-  @media #{map-get($grid-media-only, $breakpoint)} {
-    @content;
+  @if $breakpoint == 'all' {
```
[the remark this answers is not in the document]
sherakama: Works for me 👍

core/src/scss/components/card/_card--minimal.scss
JBCSU: Good catch! Thanks!

core/src/scss/components/card/_card--minimal.scss
```diff
@@ -0,0 +1,22 @@
+.su-card--minimal {
+  @include box-shadow($depth: 'flat', $stroke: 'none');
+  background-color: inherit;
+
+  .su-card__contents {
+    // override modular spacing at all breakpoints
+    @include grid-media-only(all) {
```
sherakama: Could you combine this with the horizontal grid-media-only(all) as well as it does create a lot of CSS and combining them would help optimize the output.

JBCSU: The only way I can think to "combine" them is to omit the one for `.su-card-horizontal`. Unfortunately, `.su-card--horizontal`'s padding is more specific, so it wins. So I do in fact need to specify it twice. :(

sherakama: Drat, thought I would ask anyhow.

Issue #430, opened 2019-06-11
Linked Card


buttonwillowsix:
Allow for an entire vertical and horizontal card to be a link — card container + image

For context, see notes here: 
https://docs.google.com/document/d/1BVBr71fxqSEzxP-yb3MAQ_5Dm2i9bY9KydJZfV_Aaxk/edit#

Issue #407, opened 2019-04-19
Accessibility for Card: Ensure heading level matches the heading's visual importance/level


cjwest:
**Accessibility for Card**
***Ensure heading level matches the heading's visual importance/level (Severity: 3)***

The headings are out of logical order, with h4 going before h3. Though the visual weight may match the heading level, the structure being communicated by these levels may not be clear.

[User Impact]
Screen reader users may not understand the relationships of content and be able to tell how these headings related to one another when they appear to be out of order.

[Code Reference]
```
<h4>This is a super headline</h4>
<h3>Headline goes here</h3>
```

**To Reproduce**
1. http://decanter.stanford.edu/section-components.html#kssref-components-card

**Expected behavior**
[Expected Behavior]
Ensure the heading order accurately communicates the content structure. This would likely be more easily understood with the heading order reversed or with the actual headings themselves being reversed.

[Code Example]
```
<h3>This is a super headline</h3>
<h4>Headline goes here</h4>

or

<h3>Headline goes here</h3>
<h4>This is a super headline</h4>
```

**Screenshots**

![image](https://user-images.githubusercontent.com/284440/56448378-6ab67d00-62c3-11e9-84ff-fc5080c9f27e.png)

**Additional context**
https://stanford.levelaccess.net/public/reporting/view_module.php?module_id=9631562

yvonnetangsu:
Addressing this by changing h4 to a span for the superheadline in #444

yvonnetangsu:
Also changed headline from h3 to h2

PR #444, opened 2019-06-12
Add a variant for Card where the whole card is a link and add option to turn headline into a link

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add a variant for the Card component where the whole card is a link (works for both default and horizontal versions)
- There's logic in the twig template that does not render button link nor CTA link in this variant.
- Fix accessibility issue found by Level Access where h4 (superheadline) was above the h3 (headline) by changing h4 to a span, since it isn't really semantically a headline anyway.
- Add option to turn headline into a link

# Needed By (Date)
- By end of week

# Urgency
- ?

# Steps to Test

1. Pull this branch, do `npm run build` and `npm run styleguide`.
2. Look at the `.su-card--link` variant and see if it looks and behaves ok including at mobile breakpoints. Check that the markup is semantic.
3. Check that the headline is now a link in all the variants except `.su-card--link` (we passed in the variable `card_headline_link = true` in the .json)
4. Add `"modifier_class": "su-card--horizontal su-card--link", ` to card.json, save the file and do `npm run styleguide`.
5. Look at the card component again - they should all render as horizontal cards where whole card is  link now. Check that it looks ok also. 

- Decanter

- https://github.com/SU-SWS/decanter/issues/430
- https://github.com/SU-SWS/decanter/issues/407
- #453


yvonnetangsu:
@JBCSU I merged your branch in - thanks. Anything else?

core/src/scss/utilities/placeholders/components/_card.scss
```diff
@@ -35,20 +36,22 @@
...
       @include margin(null null 0);
       @include sans;
     }
 
     // Super Headline.
-    > h4 {
-      @include modular-spacing('padding', null null -2);
+    > span {
+      @include modular-spacing('margin', null null -2);
+      @include types;
```
JBCSU: You're setting `margin-bottom` on this `span` 3 times: line 40, line 46, and in the `types` mixin on line 47. It's hard to figures out what the `margin-bottom` is supposed to be.

core/src/templates/components/card/card.json
```diff
@@ -4,7 +4,7 @@
   "card_headline": "Headline goes here",
   "card_body": "<p>Sample text goes here and this card should be able to handle a few lines.</p>",
   "card_link": "https://stanford.edu",
-  "card_cta_attributes": " rel='no-follow'",
+  "card_cta_attributes": "rel='no-follow'",
```
JBCSU: In >90% of our templates, we put a space in the `attributes`' value, and put no space between the tag and `{{ attributes }}` in the `.twig` file. I think we should stay consistent.

core/src/templates/components/link/link.twig
```diff
@@ -19,4 +19,4 @@
  *
  */
 #}
-<a{{ attributes }} href="{{ link_href }}" class="su-link {{ modifier_class }}">{{ link_content }}</a>
+<a {{ attributes }} href="{{ link_href }}" class="su-link {{ modifier_class }}">{{ link_content }}</a>
```
JBCSU: In >90% of our templates, we put a space in the `attributes`' value, and put no space between the tag and `{{ attributes }}` in the `.twig` file. I think we should stay consistent.

core/src/scss/utilities/placeholders/components/_card.scss
yvonnetangsu: The `margin-bottom: 0` was odd - I'm not even sure how it got there so I got rid of it. As for the @types - that's `margin-top: 0` and all heading styles is supposed to have that. Even though the superheadline is a span, it's supposed to take on the h6 style. It's probably a good idea since it might inherit some margin-top: from the site's styles.

core/src/templates/components/card/card.json
yvonnetangsu: Hmmmm....is there a reason why we do that for attributes? The no space thing actually screwed up things for me quite a few times especially when the component is nested. Perhaps we should always leave a space instead?

sherakama: Having no space between the html element and the `{{ attributes }}` variable in the twig file is a Drupal standard that doesn't pair well with his system or others. I am happy to add the space to the twig templates and remove the space from our data. Seems cleaner that way anyhow.

core/src/templates/components/card/card.twig
```diff
@@ -23,7 +23,19 @@
  * - card_button_label: The text label of the button component.
  */
 #}
-<div{{ attributes }} class="su-card {{ modifier_class }}">
+
+{# For variant where whole card is a link, change the outer wrapper from <div> to <a> #}
+{% if "su-card--link" in modifier_class %}
```
sherakama: This seems overly complex and is a breaking change as the wrapper element can be different. As another possible approach, could we add an optional `<a>` immediately after the wrapper div that is 100% height and width of the card?

core/src/templates/components/card/card.twig
```diff
@@ -35,7 +47,7 @@
   {# Card Super Headline #}
   {%- block block_card_super_headline %}
     {% if card_super_headline is not empty %}
-      <h4>{{ card_super_headline }}</h4>
+      <span>{{ card_super_headline }}</span>
```
sherakama: Shouldn't this still be a header and an `h2` with the appropriate `type-x` style applied?

core/src/templates/components/card/card.twig
```diff
@@ -50,31 +62,31 @@
   {{ card_body }}
 
   {# Card CTA Link #}
-  {%- if card_cta_label is not empty %}
+  {%- if card_cta_label is not empty and allow_links %}
```
sherakama: This is a nice little helper protector.

core/src/templates/components/link/link.twig
sherakama: The Drupal convention was to have no space but it is causing problems to do it that way in other systems whereas having the space causes no extra problems in Drupal. I say we add the space.

core/src/templates/components/card/card.json
yvonnetangsu: I see - didn't realize that's a Drupal thing, @sherakama 😄  Yeah, if it's not too much work for you guys, it would be nice to add the space to the attribute so it's consistent with the other variables. Thanks!

core/src/templates/components/card/card.twig
yvonnetangsu: I believe there was a discussion at some point, and the conclusion was that things that would appear in the "super headline" area probably shouldn't be treated as a HTML header element - usually things like category of the article. There's probably room for debate on this one though.

core/src/templates/components/card/card.twig
yvonnetangsu: Yeah, that's from @JBCSU 's twig tweak. I just did `if "su-card--link" is not in modifier_class` before, but this is cleaner.

core/src/templates/components/card/card.twig
yvonnetangsu: I can try out this approach.

core/src/templates/components/card/card.json
JBCSU: Yeah, that'd be great if we could put the space in the template vs. the data.

core/src/templates/components/card/card.twig
sherakama: Ok, I am fine with this, just wanted to check. Should we confirm with one other?

yvonnetangsu: Sure, perhaps @cjwest  or @buttonwillowsix?

core/src/templates/components/card/card.twig
yvonnetangsu: @sherakama , I modified it using your approach - ready for another look. Thanks! Tested on IE, Edge, and the others.

core/src/templates/components/card/card.twig
```diff
@@ -3,78 +3,94 @@
...
+{%- endif -%}
+
+<div {{ attributes }} class="su-card {{ modifier_class }}">
+  {%- if allow_links == false %}
+    <a href="{{ card_link }}" {{ card_cta_attributes }}>
   {%- endif -%}
 
-  <div class="su-card__contents">
+    {# Card Image #}
+    {% if card_image is not empty -%}
```
sherakama: Does the variable create extra whitespace? Seems a little redundant without any other markup around the variable.

yvonnetangsu: The if statement around the card_image? I think that was just there since the beginning. I'd be happy to remove it.

core/src/templates/components/card/card.twig
```diff
@@ -3,78 +3,94 @@
...
+{# For variant where whole card is a link, change the outer wrapper from <div> to <a> #}
+{% if "su-card--link" in modifier_class %}
+  {% set allow_links = false %}
+{% else %}
+  {% set allow_links = true %}
+{%- endif -%}
+
+<div {{ attributes }} class="su-card {{ modifier_class }}">
+  {%- if allow_links == false %}
+    <a href="{{ card_link }}" {{ card_cta_attributes }}>
```
sherakama: I like this approach as re-uses the variable but it has a couple of trade offs.

You can only use the outer most linking of the card if the card--link variant is passed through. I would suspect we could have more than one use case or variant that will allow for full linking and it would require us to add all of the variants to this template to make them work. What if, instead of depending on the class name to trigger the field, it was its own variable? eg: `card_link_wrapper` or something. Or perhaps, instead of the variant name check, add another boolean variable that is the toggle. eg: Add a variable that is `outer_link` and if set to true, render around, if set to false, render the button.

@kerri-augenstein Would there, should there, could there, be a use case where the whole card was wrapped and the CTA button still there? You would only be able to navigate to the outer-most-wrapper link but visually, we could still render a button without a link.

core/src/templates/components/card/card.twig
```diff
@@ -3,78 +3,94 @@
...
+        <span>{{ card_super_headline }}</span>
+      {% endif -%}
+    {% endblock -%}
 
-  {# An open field for body content #}
-  {{ card_body }}
+    {# Card Headline #}
+    {%- block block_card_headline  %}
+      {%- if card_headline is not empty %}
+        <h3>{{ card_headline }}</h3>
```
sherakama: What about making this an h2?

core/src/templates/components/card/card.twig
```diff
@@ -3,78 +3,94 @@
...
-      {
-      "link_href": card_link,
-      "link_content": card_button_label,
-      "modifier_class": "su-button",
-      }
-      only
-    -%}
-    </div>
-  {% endif -%}
+    {# Card CTA Link #}
```
sherakama: Would we still want to visually render the link or button even when the whole card is wrapped?

core/src/templates/components/link/link.twig
sherakama: 😄

core/src/templates/components/card/card.twig
sherakama: 🤔 Leave it alone for now. Thinking about this a little more, it may come in useful when we transition to a media space.

core/src/templates/components/card/card.twig
yvonnetangsu: This is probably site dependent but it seems a little more common as h3? Since a lot of page headers are h2, so card titles should be at least one step lower? One example is the Stanford homepage. I could be convinced though 😄

core/src/templates/components/card/card.twig
yvonnetangsu: @sherakama , I'm fine either way. This might be more of a @JBCSU question 🤔

core/src/templates/components/card/card.twig
yvonnetangsu: Had a discussion with @kerri-augenstein yesterday and we were ok with not rendering the a button/link when whole card is wrapped in a link - but it's nice to get another confirmation from her.

core/src/templates/components/card/card.twig
yvonnetangsu: > I like this approach as re-uses the variable but it has a couple of trade offs.
> 
> You can only use the outer most linking of the card if the card--link variant is passed through. I would suspect we could have more than one use case or variant that will allow for full linking and it would require us to add all of the variants to this template to make them work. What if, instead of depending on the class name to trigger the field, it was its own variable? eg: `card_link_wrapper` or something. Or perhaps, instead of the variant name check, add another boolean variable that is the toggle. eg: Add a variable that is `outer_link` and if set to true, render around, if set to false, render the button.
> 
> @kerri-augenstein Would there, should there, could there, be a use case where the whole card was wrapped and the CTA button still there? You would only be able to navigate to the outer-most-wrapper link but visually, we could still render a button without a link.

Actually I'll just do that - extra option to not use the class name. That will make it more flexible.

yvonnetangsu: Hm....this is more complicated than I thought, because there's some styling on su-card that needs to be removed if there is an outer_link, also for the horizontal variant, if there is an outer_link, the grid styling on the parent container needs to be removed and put on outer_link instead. Could we leave that till later?

core/src/templates/components/card/card.twig
JBCSU: `<h3>` leaves us more vulnerable to validation / accessibility warnings about skipping levels. We can be reasonably certain any page a card is on will have an `<h1>`, but it's reasonably likely that a page doesn't have an `<h2>`. Also, if we're changing the container to `<article>` per #421, `<h2>` might make more sense. So I guess I have a slight preference for `<h2>`. Which, for the record, is the opposite of my inclination when I started typing this comment. I swayed myself! 🙀

yvonnetangsu: Haha....fine. h2 it is.

sherakama:
🎉

core/src/templates/components/card/card.twig
yvonnetangsu: Headline has been changed to h2

core/src/scss/utilities/placeholders/components/_card.scss
yvonnetangsu: Removed redundant margin-bottom

core/src/templates/components/card/card.twig
yvonnetangsu: We might come back to this later and add a boolean variable.

core/src/templates/components/card/card.twig
yvonnetangsu: @cjwest , are you ok with me changing the "super headline" field from <h4> to a <span> so it doesn't cause problems with headline hierarchy (Level Access suggested that we shouldn't have a h4 above a h3)? I changed it to span since this is usually the place for metadata type info - category for example. We've also changed the headline to <h2>.

cjwest: Hi @yvonnetangsu,
Hmm.... I wonder if there is something we could use to be more semantic than <span>?

yvonnetangsu: > Hi @yvonnetangsu,
> Hmm.... I wonder if there is something we could use to be more semantic than ?

`span` is better than `div` at least 😄 , if you find anything better, let me know.

cjwest: Hi @yvonnetangsu,
I think you have it. I don't thing there's anything better than a `<span>`  or a `<p>`.
Would the `<p>`  make it easier to style?

Thanks for giving me a chance to look into it.

yvonnetangsu: Thanks, @cjwest ! Appreciate the 2nd pair of eyes. I was able to get the `span` to look ok no problem. Let's keep the `span` since it's short.

yvonnetangsu:
Since @sherakama approved, and I made all the changes that @JBCSU requested and had @caryl's confirmation about the super headline html tag, I'm going to go ahead and merge this in.

PR #449, opened 2019-06-13
305 Z-Index Map and Function

buttonwillowsix:
# READY FOR REVIEW

# Summary
- Making a z-index map and a function to easily access the map
- Obviously there is a lot more that could be done to extend this. The map provided here is based on existing elements with z-index set, but not with the same values (since 9999 should not be necessary here.)
- This PR does NOT include swapping out the existing values, but if this looks good, I could do that next.

# Needed By (Date)
- End of sprint would be nice 

# Urgency
- Not very

# Steps to Test

1. Use function z($level) somewhere and see it pull the correct value from the map or error out if you use a bad name

- JIRA ticket
- Other PRs

# See Also


core/dist/css/decanter-grid.css
```diff
@@ -1,1910 +0,0 @@
-html {
```
sherakama: I'm pretty sure we want to keep this file. Run and commit a new build.

package.json
```diff
@@ -33,7 +33,8 @@
     "url": "https://github.com/SU-SWS/decanter/issues"
   },
   "dependencies": {
-    "bourbon": "^5.1.0"
+    "bourbon": "^5.1.0",
+    "webpack-cli": "^3.3.0"
```
sherakama: This isn't really a dependency.

package.json
```diff
@@ -67,8 +68,6 @@
     "url-loader": "^1.1.2",
     "webpack": "^4.29.6",
     "webpack-assets-manifest": "^3.1.1",
-    "webpack-cli": "^3.3.0",
-    "webpack-command": "^0.4.2",
```
sherakama: Wathchew got against webpack-command?

core/src/scss/utilities/functions/z-index/_z-index.scss
```diff
@@ -0,0 +1,18 @@
...
+//
+// A function to help us get the z-index from the level from the map
+//
+// $level - the name of the z-index level
+//
+// Style guide: Functions.z-index.level
+//
+
+
+@function z($level) {
```
sherakama: Seems straightforward enough. Do you plan on adding a highest or lowest function where they return +1 above the biggest number in $z-levels or -1 of the smallest?

core/src/scss/utilities/functions/z-index/_z-index.scss
```diff
@@ -0,0 +1,18 @@
...
+  @if map-has-key($z-levels, $level) {
```
sherakama: Perhaps make the `$level` param default to `default` as that is an option in the map.

sherakama:
I think you are off to a good start. The details will be in how you implement the new map. Keep going.

core/dist/css/decanter-grid.css
buttonwillowsix: Hm, seems like my stuff did more than I wanted it to. So yes.

package.json
buttonwillowsix: It wouldn't run with webpack command. Let me try to fix!

core/src/scss/utilities/functions/z-index/_z-index.scss
buttonwillowsix: Yes, but after looking at examples of that, I didn't see one that actually seemed to work.

Issue #355, opened 2019-03-25
add same animation to more-link as exists for action-link


JBCSU:
When you hover over a `.su-link--action`, the little arrow in `:after` moves smoothly a few px to the right. I think the little double-arrow after `.su-link--more` should do the same. Should the icon after `.su-link--external` also move?

yvonnetangsu:
Yeah, good question. John was saying he had some animation in mind, e.g., the external icon would move 45 degree to the top right slightly when hovered over. He might not have a chance to document those before he left because it wasn't a priority...let me see if I can find anything about that. @josephgknox, did John mention to you and Kerri about the link icon animations? 🤔

josephgknox:
This wasn't documented, nor do I recall being part of a discussion specifically on this. I'm 100% behind the idea, though. It is something that we (or I) can bring up with Kerri, though.

PR #450, opened 2019-06-13
Link variants

JBCSU:
# READY FOR REVIEW

# Summary
- Add new link variants
    - `.su-link--download`
    - `.su-link--internal`
    - `.su-link--jump`
    - `.su-link--video`
- Change icon for `.su-link--external` per #341 
- Provide same animation for all link variants, in particular `.su-link--more` (#355)

# Needed By (Date)
- End of Decanterpalooza

# Urgency
- Minor

# Steps to Test

1. Checkout the branch
1. Rebuild the assets and the style guide
1. Test all the link variants
1. Review the doc for the new link mixins, `link-glyph` and `link-icon`

None

- #341 
- #355 

# See Also


core/src/scss/utilities/mixins/link/_link-icon.scss
```diff
@@ -0,0 +1,38 @@
...
+      mask-size: contain;
+      background-color: color(link);
+      transition: margin-left 0.2s ease-in-out;
+    }
+
+    &:hover,
+    &:focus {
+      &::after {
+        background-color: color(link--hover);
+        @if $animate {
```
yvonnetangsu: I like the mixins, @JBCSU . I think we need more options for `$animate` other than true/false though. John mentioned that the external link icon would move towards the top-right direction for example, and I imagine the download one or the jump link icon would move downward when hovered

JBCSU: I don't think that's a good idea. But if we do decide to do it, I propose we make a new ticket and do a separate PR. I don't believe that behavior was specified in the scope for the two tickets I'm addressing.

yvonnetangsu: Haha...I don't think having different animations is that bad. It would feel more natural for a down arrow to animate by moving down for example. I agree though that we can open a separate ticket for those.

core/src/scss/utilities/mixins/link/_action-link.scss
```diff
@@ -1,33 +0,0 @@
...
-// @action-link($color: $color-bright-blue)
-//
-// Provides a right carat link with alternate color.
-//
-// $color - Default color. Defaults to $color-bright-blue.
-// $hover - Hover color. Defaults to $color-black.
-//
-// Style guide: Mixins.Link.action-link
-//
-@mixin action-link($color: color(link), $hover: color(link--hover)) {
```
sherakama: Instead of completely removing the function, please deprecate it and patch through to the new option.

core/src/scss/utilities/mixins/link/_external-link.scss
```diff
@@ -1,32 +0,0 @@
...
-// @external-link($external-link, $external-link-hover)
-//
-// Default external link styling, with external icon.
-//
-// $external-link - filename of external link png & svg
-// $external-link-hover - filename of external link png & svg on hover
-//
-// Style guide: Mixins.Link.external-link
-//
-@mixin external-link($external-link: external-link, $external-link-hover: external-link-hover) {
```
sherakama: Instead of completely removing the mixin, please deprecate it and patch through to the new option.

core/src/scss/utilities/mixins/link/_more-link.scss
```diff
@@ -1,16 +0,0 @@
-@charset "UTF-8";
-
-//
-// @more-link()
-//
-// Default more link styling, with raquo.
-//
-// Style guide: Mixins.Link.more-link
-//
-@mixin more-link {
```
sherakama: Instead of completely removing the mixin, please deprecate it and patch through to the new option.

core/src/scss/utilities/mixins/link/_action-link.scss
JBCSU: Done

core/src/scss/utilities/mixins/link/_external-link.scss
JBCSU: Done

core/src/scss/utilities/mixins/link/_more-link.scss
JBCSU: Done

JBCSU:
> @JBCSU can we change the glyph used for the CTA link to use the chevron-icon instead?
> 
> From this:
> cta-text_with_glyph
> 
> To this:
> Screen Shot 2019-06-14 at 2 41 04 PM

@kerri-augenstein Done.

core/src/scss/utilities/mixins/link/_action-link.scss
```diff
@@ -1,33 +1,22 @@
...
-    position: relative;
-    transition: left 0.2s ease-in-out;
-    text-decoration: none;
-  }
-
-  &:hover {
-    color: $hover;
+  $msg: "The action-link mixin has been deprecated and will be removed in ";
+  $msg: $msg + "version 6. Use link-icon(caret-right, 0.5em) instead.";
+  @warn $msg;
```
sherakama: :heart: this. Thank you.

core/src/scss/utilities/mixins/link/_link-glyph.scss
```diff
@@ -0,0 +1,30 @@
...
+    content: '#{$glyph}';
+    @include margin(null null -1px 0.4em);
+    transition: margin-left 0.2s ease-in-out;
+  }
+
+  &:hover,
+  &:focus {
+    &::after {
+      @if $animate {
+        @include margin(null null null 0.55em);
```
sherakama: instead of using margin to do the animation, try using `translateX()` as it does not affect the space around it and it will not cause reflow or shifting of other elements.

JBCSU: Good tip. Thanks! This is done now.

sherakama:
Ticket created: https://github.com/SU-SWS/decanter/issues/456

PR #454, opened 2019-06-14
305: Z-Index Map and Function

buttonwillowsix:
# READY FOR REVIEW

# Summary
- Making a z-index map and a function to easily access the map
- Obviously there is a lot more that could be done to extend this. The map provided here is based on existing elements with z-index set, but not with the same values (since 9999 should not be necessary here.)
- This PR does NOT include swapping out the existing values, but if this looks good, I could do that next.

# Needed By (Date)
- End of sprint would be nice, from a personal standpoint. :)

# Urgency
- Not very

# Steps to Test

1. Use function z($level) somewhere and see it pull the correct value from the map or error out if you use a bad name

- JIRA ticket
- Other PRs

# See Also


core/src/scss/utilities/variables/core/z-index.scss
```diff
@@ -0,0 +1,34 @@
...
+// Style guide: Variables.Core.Z-index.level
+//
+$z-levels: (
+  "navigation":  100,
+  "button":      100,
+  "masthead":     20,
+  "overlay":      10,
+  "default":       1,
+  "background":  -100
+);
```
yvonnetangsu: Are we adding the main nav z-indexes in here? Another good one to add might be the skiplinks (focus state) since that needs to be above everything in the masthead when it comes into focus. Looks like a great start 😄 👍 - thanks for doing this much needed wrok!

yvonnetangsu: Are we adding the main nav z-indexes in here? Another good one to add might be the skiplinks (focus state) since that needs to be above everything in the masthead when it comes into focus. Looks like a great start 😄 👍 - thanks for doing this much needed work!

core/src/scss/utilities/functions/z-index/_z-index.scss
```diff
@@ -0,0 +1,18 @@
...
+//
+// A function to help us get the z-index from the level from the map
+//
+// $level - the name of the z-index level
+//
+// Style guide: Functions.z-index.level
+//
+
+@function z($level) {
+  $level: "default";
```
sherakama: This no worky. This will set the passed in parameter to "default" on every call. What you are looking for is:
`@function z($level: 'default') {`

sherakama:
Please also do the swap out of the old variables in this PR.

core/src/scss/utilities/variables/core/z-index.scss
buttonwillowsix: @yvonnetangsu we could! I was wondering how many we wanted to call out explicitly, and how many we might want to make relative (so for example z-index: z(navigation) - 1). I can see benefits to either approach.

core/src/scss/utilities/functions/z-index/_z-index.scss
buttonwillowsix: Ooh! Thanks!

core/src/scss/utilities/variables/core/z-index.scss
yvonnetangsu: > @yvonnetangsu we could! I was wondering how many we wanted to call out explicitly, and how many we might want to make relative (so for example z-index: z(navigation) - 1). I can see benefits to either approach.

Yeah, that's a good question. It might be nice to have one on the map that's really high up - like main nav uses 9999 as default I believe, since we want it above most everything else. The drop down menus on the main nav has higher z-index, something like 11111 because in case someone put another main nav right under, we want the expanded drop down for the top main nav to cover up the bottom main nav, and then the lockup within the masthead has some value in between the two. So maybe main nav 9999 is another good one to be on the map - we can do relative values for the drop down and the lockup then.

yvonnetangsu: And most of the time when I try to assign z-index values to two items, I set the numbers at least 10 apart in case I decide to add something else later that's in between.

buttonwillowsix: So in thinking about this some more, it feels like the landmarks are the stuff we know we have (main nav, drop-downs, masthead, etc) and that if we define those relative to each other, and far enough apart to allow for multiple relative values to be set, people will have the ability to easily set values for things they add in (like a second main nav, etc.) The changes in my next push will reflect that approach.

buttonwillowsix:
This is ready for review again. I added another landmark between the default and the background because I felt like it needed one, but I am not attached to it. I left lots of space between things. If anyone feels like more is needed, that is OK too (I wondered about a "modal" variable.)

buttonwillowsix:
Readjusting everything except the skiplinks to account for the fact that CMS admin bars need to not be trounced. D8 is z-index: 502, WP seems to be 9999, so that is easier to avoid. Still, even with <500, there is a lot of space to move around in.

buttonwillowsix:
lockup:
z-index: 10010;

PR #461, opened 2019-06-25
Dec 71 styles switch default

yvonnetangsu:
# NOT READY

# Summary
- TL;DR - what's this PR for?

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


core/src/templates/components/main-nav/main-nav.json
```diff
@@ -8,9 +8,7 @@
     {
       "href": "#",
       "text": "First level item",
-      "attributes": {
-        "class": "su-secondary-nav__item--expanded"
-      },
+      "current": true,
```
sherakama: Oh, I meant to revert this back to the original json. The new schema here is for testing.

yvonnetangsu:
@sherakama I still have quite a bit left, but the styles are getting a little closer. Just realize can't have `@extend` inside media queries so I need to be careful 🤔

yvonnetangsu:
@sherakama, I consolidate the color maps and change the color key names. Instead of main-nav-colors and secondary-nav-colors, there's only one nav-colors now. I use the "`h-`" prefix to indicate those are for horizontal menu items (Main Nav desktop 1st level). I removed the "`mobile-`" in the color key names also. So what used to be e.g., `bg--mobile` has become just `bg` and applies to all the "vertical" menus - Main Nav mobile version, Main Nav drop down, and Secondary Nav (all).

sherakama:
Wonderful, I'll merge it down and you can keep going on this branch.

yvonnetangsu:
@sherakama I think the main nav styles are same as the one on master now. Give it a try. I gave the borders slightly different treatment compared to the Figma - now the borders have same distance to the text no matter how many levels deep they are, while in Figma, all the borders line up on the edge at the same place. I'm having a meeting with @kerri-augenstein tomorrow so we might have time to talk about this.

yvonnetangsu:
@sherakama , @JBCSU would like the placeholder/styles merged into master soon so we can use them in Redwood. Could I start a branch off of master, and pick out only the placeholder/style/colormap related changes into that branch and create a separate PR so we can merge that in before you finish the work on the secondary nav/js? 😃

PR #473, opened 2019-08-01
Add focus outline back to default buttons

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Put focus outline back on default buttons and let developers remove them in their own projects if needed, since it's really difficult to put outline back once it's been removed (no good CSS code to reset it). Having the outline as default helps with accessibility.
- Also remove extra bottom margin on some of the CTA variants.

# Needed By (Date)
- Whenever

# Urgency
- N/A

# Steps to Test

2. Go to the Components -> button section and tab through the buttons, check that the browser default outline ring is back when button is on focus. 
3. Look at the Site Search component and check that the magnifying glass icon still has outline when tab to it.
4. Check Main Nav and Masthead component at MD breakpoint or smaller. See that the hamburger menu toggle button has no outline on focus (this is an exception case that has no focus outline since it has the thick border underneath already).

- Decanter
- Redwood WP theme

- @JBCSU

### DIFF
--- a/core/src/scss/elements/input/_input.scss
+++ b/core/src/scss/elements/input/_input.scss
@@ -8,7 +8,8 @@
 input,
 textarea,
 select {
-  appearance: none;
+  -moz-appearance: none;
+  -webkit-appearance: none;
   border: $input-border-width solid $color-driftwood;
   border-radius: 0;
   box-sizing: border-box;
@@ -41,9 +42,6 @@ textarea {
 }
 
 select {
-  -moz-appearance: none;
-  -webkit-appearance: none;
-  appearance: none;
   background-color: $color-white;
   background: url("#{$image-path}/arrow-down.png");
   background-image: url("#{$image-path}/arrow-down.svg");
--- a/core/src/scss/elements/input/_input.scss
+++ b/core/src/scss/elements/input/_input.scss
@@ -8,8 +8,6 @@
 input,
 textarea,
 select {
-  -moz-appearance: none;
-  -webkit-appearance: none;
   border: $input-border-width solid $color-driftwood;
   border-radius: 0;
   box-sizing: border-box;
--- a/core/src/scss/utilities/placeholders/components/_cta.scss
+++ b/core/src/scss/utilities/placeholders/components/_cta.scss
@@ -22,6 +22,7 @@
 
   .su-cta__button {
     @include button-big;
+    @include margin(null null 0);
     @include grid-media('xs') {
       width: 100%;
       z-index: 100;
--- a/core/dist/css/decanter.css
+++ b/core/dist/css/decanter.css
@@ -319,9 +319,6 @@ button,
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   font-family: "Source Sans Pro", "Helvetica Neue", "Helvetica", "Arial", sans-serif;
-  -webkit-appearance: none;
-     -moz-appearance: none;
-          appearance: none;
   cursor: pointer;
   display: inline-block;
   border: 0;
@@ -2658,8 +2655,6 @@ img {
 input,
 textarea,
 select {
-  -moz-appearance: none;
-  -webkit-appearance: none;
   border: 0.1rem solid #b6b1a9;
   border-radius: 0;
   -webkit-box-sizing: border-box;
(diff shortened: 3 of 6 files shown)


core/src/scss/elements/input/_input.scss
```diff
@@ -8,7 +8,8 @@
 input,
 textarea,
 select {
-  appearance: none;
+  -moz-appearance: none;
```
sherakama: We shouldn't need vendor prefixes as the autoprefixer should take care of this for us.

yvonnetangsu: Oh @sherakama , this is the weird one, the one without prefix is invalid, so the only forms that are actually valid are the prefix ones so I thought I'd remove the un-prefixed one.

JBCSU: This comment from https://developer.mozilla.org/en-US/docs/Web/CSS/appearance concerns me:

> **Compatibility note**: If you wish to use this property on websites, you should test it very carefully. Although it is supported in most modern browsers, its implementation varies widely. In older browsers, even the keyword none does not have the same effect on all form elements across different browsers, and some do not support it at all. The differences are smaller in the newest browsers.

Should we be using `appearance` at all?

core/src/scss/utilities/placeholders/components/_cta.scss
```diff
@@ -22,6 +22,7 @@
 
   .su-cta__button {
     @include button-big;
+    @include margin(null null 0);
```
JBCSU: Why was this necessary? Just curious.

yvonnetangsu: In the current one, when you tab to the CTA, in a couple of them, there's a white space the whole CTA (more obvious when you tab to it so you see the focus outline).

core/src/scss/elements/input/_input.scss
yvonnetangsu: Actually, yeah, @JBCSU , I'm not really sure we need it for the browsers that we're supporting (2 versions back). @sherakama , what do you think? Could we get rid of appearance?

sherakama: +1 to removing it.

core/src/scss/utilities/placeholders/components/_cta.scss
JBCSU: Ok, I see why you want to remove the bottom margin on the CTA button. But this has the effect of changing the vertical spacing of the entire CTA. Should we not just move the bottom margin to the whole CTA, vs. eliminating it?

core/src/scss/elements/input/_input.scss
JBCSU: +1 also. Please remove it.

core/src/scss/utilities/placeholders/components/_cta.scss
yvonnetangsu: I believe we don't actually want any external spacing that's outside any Decanter elements (no spacing outside the top level wrapper) so I'm removing it all together.

core/src/scss/elements/input/_input.scss
yvonnetangsu: Gladly 😄

core/src/scss/utilities/placeholders/components/_cta.scss
JBCSU: Hmmm. Yes, I agree with that in principle. However, the fact is we did effectively have external spacing on the CTA component. If people upgrade to the latest master, their layout will be affected if they use CTA's. Maybe that's ok, but I want to make a conscious decision about it.

@sherakama, your thoughts? Are CTA's used on any of your decanter-based sites? Is SWS okay with the bottom margin going away?

core/src/scss/elements/input/_input.scss
yvonnetangsu: Ok, it's gone from both input and button-base

core/src/scss/utilities/placeholders/components/_cta.scss
yvonnetangsu: That's a good point about backward compatibility 😄

sherakama: CSS can change and the styles applied to elements can change, esp, if the new code keeps the end result the same. Keep in mind, we try to avoid top margin and try to only apply margin to the bottom of elements to stay consistent.

PR #474, opened 2019-08-02
Alert Component with html5

cjwest:
# READY FOR REVIEW
Should this go into V6 or master?

# Summary
- Proposing HTML5 syntax and role=alert to the alert component

# Needed By (Date)

# Urgency
- not critical

# Steps to Test

1. Check out this branch
1. Review and comment on changes
2. Verify the alert component works as expected.

- Decanter

- JIRA ticket - https://stanfordits.atlassian.net/browse/THEME-15
- Related to https://github.com/SU-SWS/decanter/pull/421

# See Also

### DIFF
--- a/core/src/templates/components/alert/alert.twig
+++ b/core/src/templates/components/alert/alert.twig
@@ -19,13 +19,13 @@
  * - alert_body: Text details of the alert message.
  */
 #}
-<div {{ attributes }} class="su-alert {{ modifier_class }}" >
-  <div class="su-alert__body">
+<article {{ attributes }} class="su-alert {{ modifier_class }}" role="alert" >
+  <section class="su-alert__body">
     {% if alert_header is not empty %}
     <h3 class="su-alert__heading">{{ alert_header }}</h3>
     {% endif %}
     {% if alert_text is not empty %}
     <p class="su-alert__text">{{ alert_text }}</p>
     {% endif %}
-  </div>
-</div>
+  </section>
+</article>


core/src/templates/components/alert/alert.twig
```diff
@@ -19,13 +19,13 @@
  * - alert_body: Text details of the alert message.
  */
 #}
-<div {{ attributes }} class="su-alert {{ modifier_class }}" >
-  <div class="su-alert__body">
```
cjwest: I'm curious as to why there's two levels of <div> would one be sufficient?

core/src/templates/components/alert/alert.twig
```diff
@@ -19,13 +19,13 @@
...
     {% if alert_header is not empty %}
     <h3 class="su-alert__heading">{{ alert_header }}</h3>
     {% endif %}
     {% if alert_text is not empty %}
     <p class="su-alert__text">{{ alert_text }}</p>
     {% endif %}
-  </div>
-</div>
+  </section>
+</article>
```
cjwest: @yvonnetangsu, 
I'd love your $0.02 on this.

yvonnetangsu: Will do after next week :D Rushing to the airport!

core/src/templates/components/alert/alert.twig
sherakama: Good question. There may have been a reason for this but it doesn't come to mind. I would have to investigate further. Can you provide a recommendation? Assume it was an oversight/overbuild.

cjwest: On further inspection, it looks like the nested `<div>s` are to help with layout.

sherakama:
@cjwest there seems to be a bunch of other code in this pr regarding the date-stacked component. Can you clean it up pls.

core/src/templates/components/alert/alert.twig
sherakama: @cjwest what would make this an article?

cjwest:
@sherakama,

Hmmm, all the extra files such as the the date-stacked happened when I switched to merge with V6 rather than master. Before I spend any time on a merge with V6, please let me know: Should this merge with V6 or master?

Thank you for asking about the `<article>` element. It may not be appropriate. What I was thinking was that an alert is a message is generally independent of the text on a page. That is:

> The HTML <article> element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication).

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

The alert is
1. Self contained
2. reusable

It is not intended for syndication, so that might be the show stopper. In that case, I wonder if a `<p>` would be more appropriate. 

I have not found any examples of using HTML5 for this kind of use case. Maybe I'm searching the wrong terms?

I welcome your thoughts.

core/src/templates/components/alert/alert.twig
cjwest: Thank you for asking about the `<article>` element. It may not be appropriate. What I was thinking was that an alert is a message is generally independent of the text on a page. That is:

> The HTML `<article>` element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication).

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

The alert is
1. Self contained
2. reusable

It is not intended for syndication, so that might be the show stopper. In that case, I wonder if a `<div>` would probably be more appropriate. 

I have not found any examples of using HTML5 for this kind of use case. Maybe I'm searching the wrong terms?

I welcome your thoughts.

sherakama: Thanks for the explanation, I think this all makes sense. Good to have it documented here as well.

sherakama:
@cjwest 

As this changes the outermost wrapper for the component, please change this PR to go in to `v6` and add this change to the changelog.txt file. 

Thank you.

core/src/templates/components/alert/alert.twig
yvonnetangsu: Hmmm....I prefer `<section>` over `<article>` myself, since an alert sometimes only make sense when it's shown together with some other elements on a page (eg, alert of a successful submission) so it's not quite independently distributable in that case 😃

sherakama: Perhaps `<div>` from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

cjwest: I don't think `<section>` would be appropriate:
> A rule of thumb is that a section should logically appear in the outline of a document.

I think either `<div>` or `<article>` would be appropriate. I'm still leaning towards `<article>`. With regard to the alert being related to a successful submission, I think the submission is as not _necessarily_ related to a page. That is, a submission is with regard to an _action_, not really the content. 

I welcome your thoughts, @yvonnetangsu and @sherakama.

cjwest:
Closing in lieu of:
https://github.com/SU-SWS/decanter/pull/481

Issue #475, opened 2019-08-02
Split input.scss into different files 


yvonnetangsu:
**Is your feature request related to a problem? Please describe.**
Our input.scss has a lot of custom styles for radio buttons, checkboxes and range selectors. Some of which might not be needed for projects.

**Describe the solution you'd like**
Split the input.scss file into several smaller ones with styles for different types of input elements.

yvonnetangsu:
Merged

PR #476, opened 2019-08-02
Split input.scss into several files and add new input-base placeholder

yvonnetangsu:
# READY FOR REVIEW

# Summary
- The current input.scss file inside the elements directory has styling for many input elements that might not be used for every project (eg, range, custom radio button/check boxes styling). This PR splits it up into several files so developers can choose to cherry pick the elements they need. 
- Added input-base placeholder.
- Removed "role=navigation" for Global Footer nav (Level Access recommendation).
- Note: styling for radio button/checkbox/range probably need work later, but the purpose of this PR isn't to change them so no work has been done to them.

# Needed By (Date)
- Whenever

# Urgency
- N/A

# Steps to Test

2. Check that everything that has input elements still functions the same.

- Decanter
- Redwood WP theme

- #475 
- @JBCSU


yvonnetangsu:
I'm back :D @sherakama and @JBCSU , if you have time, could you please take a quick look at this one? And feel free to @ me if there are other PR waiting for my review. Thanks!!

core/src/scss/elements/input/_radio-checkbox.scss
```diff
@@ -0,0 +1,85 @@
+@charset "UTF-8";
+
+// Custom checkboxes and radio buttons.
+[type='checkbox'],
+[type='radio'] {
+  // The actual input element is only visible to screen readers, because
+  // all visual styling is done via the label.
+  @include hide-visually;
+
+  .lt-ie9 & {
```
sherakama: Where does this class come from?

core/src/scss/elements/input/_select.scss
```diff
@@ -0,0 +1,14 @@
+@charset "UTF-8";
+
+// Select dropdown
+
+select {
+  @extend %input-base;
+  outline: none;
```
sherakama: Does this affect the focus state? I don't think we want to remove the outline on the focus of any input without handling it a different way for visually impaired users.

core/src/scss/elements/input/_radio-checkbox.scss
```diff
@@ -0,0 +1,85 @@
...
+  box-shadow: 0 0 0 1px $color-driftwood;
+  content: '\a0';
+  display: inline-block;
+  height: 1.8rem;
+  line-height: 1.8rem;
+  margin-right: 0.6em;
+  text-indent: 0.15em;
+  vertical-align: middle;
+
+  // Target IE 11 and below to vertically center inputs.
```
sherakama: I don't see why the width would be for IE 11 and below for vertical center. Can you elaborate on this comment?

core/src/scss/elements/input/_range.scss
```diff
@@ -0,0 +1,92 @@
...
+[type='range'] {
+  -webkit-appearance: none;
+  border: 0;
+  padding-left: 0;
+  width: 100%;
+}
+
+[type='range']:focus {
+  box-shadow: none;
+  outline: none;
```
sherakama: Just want to double-check that this is still ok. I know this was from the previous state but I want to double-check that we haven't learned more since it was initially committed.

core/src/scss/elements/input/_radio-checkbox.scss
yvonnetangsu: @sherakama , I actually didn't do any work to "improve" some of the current input styles nor look at them that closely since the purpose of the PR is just to split the input.scss file into smaller chunks so developers can cherry pick only the parts they need for their projects. I agree that many things need work (quite a bit in fact) for our current Decanter input styles, but I'd like to save that for another PR. Let me know what you think.

core/src/scss/elements/input/_select.scss
yvonnetangsu: I actually would vote for putting the outline back. However, it was in the original input.scss and might be used by some sites so I kept the `outline: none;` but took it out of the placeholder. What do you think? Should I just put the outline back into everything? @JBCSU  please feel free to comment too.

core/src/scss/elements/input/_radio-checkbox.scss
yvonnetangsu: Again, not my code - I just copied and pasted them into their own file 😂 . The radio button/ checkboxes actually need a ton of work. There is a lot of questionable stuff there.

core/src/scss/elements/input/_range.scss
yvonnetangsu: Please see above.

core/src/scss/elements/input/_select.scss
yvonnetangsu: > Does this affect the focus state? I don't think we want to remove the outline on the focus of any input without handling it a different way for visually impaired users.

And yes, it does affect the focus state. The current Decanter style remove all outline for focus state, and gives them a drop shadow instead which I'm not crazy about either.

PR #481, opened 2019-08-21
THEME-15: Proposing to use article and section elements, and role="alert"

cjwest:
…ert"

# READY FOR REVIEW

# Summary
- Replaces https://github.com/SU-SWS/decanter/pull/474
- Please see conversation at https://github.com/SU-SWS/decanter/pull/474
- Proposing HTML5 syntax and role=alert to the alert component

# Needed By (Date)

# Urgency
- Not Urgent

# Steps to Test

1. Check out this branch
2. Review and comment on changes
3. Verify the alert component works as expected.

- Decanter

- JIRA ticket- https://stanfordits.atlassian.net/browse/THEME-15
- Other PRs: This takes the place of https://github.com/SU-SWS/decanter/pull/474
- I rebuilt this PR to merge into V6 instead of Master. 
- Anyone who should be notified? 
@sherakama 
@yvonnetangsu 
@rebeccahongsf 

# See Also

### DIFF
--- a/core/src/templates/components/alert/alert.twig
+++ b/core/src/templates/components/alert/alert.twig
@@ -19,13 +19,13 @@
  * - alert_body: Text details of the alert message.
  */
 #}
-<div {{ attributes }} class="su-alert {{ modifier_class }}" >
-  <div class="su-alert__body">
+<article {{ attributes }} class="su-alert {{ modifier_class }}" role="alert">
+  <section class="su-alert__body">
     {% if alert_header is not empty %}
     <h3 class="su-alert__heading">{{ alert_header }}</h3>
     {% endif %}
     {% if alert_text is not empty %}
     <p class="su-alert__text">{{ alert_text }}</p>
     {% endif %}
-  </div>
-</div>
+  </section>
+</article>


cjwest:
I don't think `<section>` would be appropriate:
> A rule of thumb is that a section should logically appear in the outline of a document.
From: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

I think either `<div>` or `<article>` would be appropriate. I'm still leaning towards `<article>`. With regard to the alert being related to a successful submission, I think the submission is as not _necessarily_ related to a page. That is, a submission is with regard to an _action_, not really the content. 

I welcome your thoughts, @yvonnetangsu and @sherakama.

core/src/templates/components/alert/alert.twig
```diff
@@ -19,13 +19,13 @@
  * - alert_body: Text details of the alert message.
  */
 #}
-<div {{ attributes }} class="su-alert {{ modifier_class }}" >
-  <div class="su-alert__body">
+<article {{ attributes }} class="su-alert {{ modifier_class }}" role="alert">
```
yvonnetangsu: Hi @cjwest , unfortunately I don't believe we can use `role='alert'` here. It implies that the element is a "live region" so more appropriate for dynamic alerts, which won't work well for a news type alert, e.g., "extreme heat on campus". As for the outside wrapper, I'm happy to change the `<section>` to a more generic `<div>`. Perhaps ask JB's opinion about whether to use article or div. She usually has good argument for that type of things :).

cjwest: Thank you @yvonnetangsu !!!

@JBCSU,
I'd love your $0.02 as I'm learning about how to use HTML5. 
What do you think, should we use `<article>` or `<div>` to wrap the alert component?

cjwest: This is new. I suddenly can't push my next commit. I get the message:
```
remote: Resolving deltas: 100% (7/7), completed with 5 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/v6-components-html5-alert.
remote: error: At least 1 approving review is required by reviewers with write access.
To github.com:SU-SWS/decanter.git
 ! [remote rejected] v6-components-html5-alert -> v6-components-html5-alert (protected branch hook declined)
error: failed to push some refs to 'git@github.com:SU-SWS/decanter.git'
```

JBCSU: I do not think `<article>` is appropriate for an alert. An `<article>` should make sense in an RSS feed; I don't believe alerts would makes sense in that context.

`<aside>` is another option, but I think `<div>` is fine.

cjwest:
Closing this because the branch name starts with "v"

Please see:
https://github.com/SU-SWS/decanter/pull/482

Issue #456, opened 2019-06-17
Add option to fine tune position of icon in link


sherakama:
See: https://github.com/SU-SWS/decanter/pull/450#pullrequestreview-250166249

yvonnetangsu:
Done

PR #504, opened 2019-10-02
Add parameter to @link-icon to adjust vertical position of icon and different options for $animate parameter

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add a parameter `$vertical` in the `@link-icon` mixin so users can adjust the vertical position of the icon by adding a corresponding `margin-bottom` to the `::after` pseudo element of the link. Positive value moves icon up; negative moves it down.
- Add options for `$animate: right, up, down, topright` (diagonally towards upper right corner).
- Finetune icon vertical positions of existing link component variants.

# Needed By (Date)
- Whenever

# Urgency
- Not urgent

# Steps to Test

1. Pull this branch and compile styleguide.
2. Go to the compiled styleguide, look at Components -> Links, check that all the icons of existing link variants have good vertical alignments with the link text and that they animate as specified in #505 
3. Open a link variant scss file, e.g., components/link/_link--internal.scss, change the `$vertical `and `$animate` parameters and check that they behave as expected.

- Decanter and any sites that use link variants that uses SVG masks

- #456 
- #505


core/src/scss/utilities/mixins/link/_link-icon.scss
```diff
@@ -1,21 +1,22 @@
...
 //
 // $icon {string} - Basename of link icon .svg file (without the .svg suffix).
 // $width {string} - Width of icon including unit, e.g. 0.75em, 24px.
+// $vertical {string] - Adjustment of vertical position of icon, e.g., '2px' moves icon up 2 pixels, '-0.3em' moves it down 0.3 em.
 // $animate {boolean} - Whether or not to animate on hover.
 //
 // Style guide: Mixins.Link.link-icon
 //
-@mixin link-icon($icon, $width: 0.65em, $animate: true) {
+@mixin link-icon($icon, $width: 0.65em, $vertical: null, $animate: true) {
```
sherakama: Technically, changing the order of the parameter input is changing API so in order to not break API, you will need to add it to the end of the mixin parameters. It is much cleaner where it is right now, so if you would like to put this in as a PR to `v6` I'll be ok with where it is. 

Also, shouldn't the default value for the vertical alignment be `-1px` as that was the original value?

yvonnetangsu: Sounds good, @sherakama, and thanks for reviewing. When is v6 coming out? Is it after Decanterpalooza 3 this Dec? If so, then it's totally fine that we put this in v6. I just wanted to do this before we work on those new card/hero variants so that the icons are in the right place for those CTA links.

I actually found that a few of those icons don't need any vertical adjustments/margin-bottom so I'd rather have the default be `null` instead. Some of those with the `-1px` value look slightly misaligned.

yvonnetangsu: Actually, @sherakama, now that I think about it, I'm going to change the $animate parameter since Kerri asked for a few different animation. In that case, maybe I'll keep the original parameter order 🤔

sherakama: Yeah, V6 is planned for release after the next Palooza sprint which is in December. We will likely release a new version in January as we usually have a few items to wrap up after.

yvonnetangsu:
@sherakama , I added options to `$animate` per Kerri's request and swapped the order of `$animate` and `$vertical` so `$animate` is back to being the 3rd parameter. Could you please take another look? Is this still API breaking?

core/src/scss/components/link/_link--internal.scss
```diff
@@ -7,5 +7,12 @@
 // @see scss/utilities/mixins/link/_link-icon.scss
 //
 .su-link--internal {
-  @include link-icon(lock, 0.75em);
+  @include link-icon(lock, 0.75em, none, -0.06em);
+
+  &:hover,
+  &:focus {
+    &::after {
+      mask-image: url("#{$su-image-path}/lock-solid.svg");
```
sherakama: Support looks good for this and the autoprefixer rendered the right prefixes.
https://caniuse.com/#search=mask-image

Issue #598, opened 2019-12-19
Replace "@elseif" with "@else if" in Sass code


nrambeck:
**Describe the bug**

The latest version of Dart Sass generates a deprecation warning when it finds `@elseif` in the Decanter code. Evidently `@elseif`will be removed in the future, in favor of the primary else syntax of `@else if`. 

```
WARNING on line 73, column 7 of node_modules/decanter/core/src/scss/utilities/mixins/animation/_fancy-hover.scss: 
@elseif is deprecated and will not be supported in future Sass versions.
Use "@else if" instead.
   ╷
73 │     } @elseif ($position == 'top') {
   │       ^^^^^^^
   ╵
```

**To Reproduce**
1. Install Dart Sass version 1.23.x
2. Compile Decanter Sass with Dart Sass
3. You should see a warning similar to the one in the description above

**Expected behavior**

Expect to have no deprecation warnings with the latest version of Dart Sass.

**Additional context**

This was discovered while working on new themes for the H&S Drupal platform. We want to use Dart Sass to leverage some of the latest features Sass has to offer. Our team is happy to create a pull request for this work, if desired.

More info about the deprecation change: https://github.com/sass/dart-sass/pull/317

sherakama:
Thanks for this. We're also seeing it in the latest sass compiler as well. Definitely will do.

sherakama:
PRs always welcome 😉

PR #600, opened 2019-12-20
Fix deprecated '@elseif' syntax

nrambeck:
# READY FOR REVIEW

# Summary

Fixes a few instances in Sass of `@elseif` syntax and replaces it with the support `@else if` syntax. Newer versions of Sass have deprecated the `@elseif` syntax and it will be dropped in future versions.

# Needed By (Date)

No specific date

# Urgency

Nothing is broken now. For those using the latest version of Sass, there are deprecation warning in every build which can cause lots of noise and make it hards to see other errors. The old syntax will probably continue to be supported for 6 months to a year or more.

# Steps to Test

1. Compile Decanter (before this commit) with the latest version of Dart Sass
1. You should see deprecation warnings
1. Compile this PR with the latest version of Dart Sass
1. The deprecation warnings should no longer be generated

The H&S Drupal 8 project is currently getting these deprecation errors with new themes that are currently in development. This is because we are using the latest version of Dart Sass.

- SU-SWS/decanter#598
- @sherakama, @silverli 

# Accessibility Checks

- None, N/A

# See Also

### DIFF
--- a/core/src/scss/utilities/mixins/animation/_fancy-hover.scss
+++ b/core/src/scss/utilities/mixins/animation/_fancy-hover.scss
@@ -70,7 +70,7 @@
       &::before {
         bottom: 0;
       }
-    } @elseif ($position == 'top') {
+    } @else if ($position == 'top') {
       @include padding(0.8em null null);
 
       &::before {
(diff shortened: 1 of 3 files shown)


cjwest:
Thank you, @nrambeck !

PR #619, opened 2020-02-12
feat: add calculate-font-size function

heatherdesigns:
# READY FOR REVIEW

# Summary
We would like to use Decanter sass styles and HTML for the immense time and effort it will save us, and for the branded consistency it provides, especially in places like the Brand bar, navigation, footer, and lockups.

Currently, we have our own SCSS styles set with a base font-size set on the html element of 16px. When trying to use Decanter for our main navigation styles, we realized that the Decanter base font-size is set to 62.5%, or what comes out to 10px in those styles.

I think we are open to using 10px as the base font size for our styles if it is necessary, or if the Decanter team has a reason they prefer that base font-size, and we could then adjust our own styles and where we use rem values currently. However, I did want to mention that 16px as a base font-size is quite standard across the internet and different browsers, and this compelling article also has SEO reasons listed for keeping it 16px as well: https://www.mediavine.com/increase-font-size-and-increase-seo-rpm-along-with-it/

I did see that the `$su-root-font-size` is a variable that can be adjusted, which I was excited about https://github.com/SU-SWS/decanter/blob/master/core/src/scss/elements/typography/_typography.scss#L29:L32 ! However, if I adjust the `$su-root-font-size` variable to 16px, then because rems were used in Decanter styles based on the assumption that base font size was 10px, everything on our site using Decanter styles becomes huge.

For example, the Decanter style for the main navigation li a font-size is set to 2.1rems, which is 21px on your site, but becomes 33.6px on our site.

I was hoping that the Decanter design system might be a bit more flexible, and build in some ways that base font-size could be set alternatively, to accommodate for the different ways projects might write their styles.
 
## A potential alternative solution
My coworker Heather and I had one idea we wanted to run by you that could allow rems to continue to be used in both Decanter as they are now, and in our styles as well (as we understand rems are very accessible for websites to resize text).

We came up with a `calculate-font-size` function that could be used in Decanter such that your team could continue using 10px as a base font-size, but it would not drastically impact styles on another site that might not use that as the base font size, but maybe used 16px, or 18px, etc.

Please take a look at the files changed in this PR to see the function, some comments about the function, and it's potential implementation in one place in the Decanter code base. If you are happy with the solution, we can make the changes everywhere, but we didn't want to do that without chatting with you first!

We also made a Codepen if you'd rather play around with the function we made in that environment instead! https://codepen.io/silverli/pen/rNVOajq?editors=1100 

## Expected Behavior
We were hoping for a bit more flexibility with Decanter for setting CSS values, including using a different base html font-size than 10px.

# Needed By (Date)
No specific date

# Urgency
We are working on brand-bar, footer, and navigation elements this week, so we would like to know a best way forward for now, or what solution might align with your goals.

# Steps to Test
Try switching `$su-root-font-size` to 16px and see that styles across Decanter look very large.

The H&S Drupal 8 project is currently seeing a very large main navigation when we pull in the Decanter styles. This is because our base-font size and rem values we use don't align with Decanter's.

[@silverli](https://github.com/silverli), [@heatherdesigns](https://github.com/heatherdesigns)

### DIFF
--- a/core/src/scss/utilities/variables/core/_typography.scss
+++ b/core/src/scss/utilities/variables/core/_typography.scss
@@ -12,7 +12,7 @@
 /// @name $su-root-font-size
 ///
 /// @group variable
-$su-root-font-size: 62.5%;
+$su-root-font-size: 10px;
 
 ///
 /// Font size.
--- a/core/src/scss/components/alert/_alert.scss
+++ b/core/src/scss/components/alert/_alert.scss
@@ -17,36 +17,36 @@
 // Style guide: Simple.Alert
 //
 .su-alert {
-  @include padding(2rem);
+  @include padding(calculate-font-size(20px));
   background-color: color(default, $su-alert-colors);
-  background-position: 1rem center;
+  background-position: calculate-font-size(10px) center;
   background-repeat: no-repeat;
-  background-size: 4rem;
+  background-size: calculate-font-size(40px);
   color: color(dark-text, $su-alert-colors);
 
   @include grid-media('md') {
-    background-size: 5.2rem;
+    background-size: calculate-font-size(52px);
   }
 
   .su-alert__body {
     display: table-cell;
     vertical-align: top;
 
-    @include padding(null null null 3.5rem);
+    @include padding(null null null calculate-font-size(35px));
 
     @include grid-media('md') {
-      @include padding(null null null 5rem);
+      @include padding(null null null calculate-font-size(50px));
     }
 
     p:last-child {
-      @include margin(null null 0.8rem);
+      @include margin(null null calculate-font-size(8px));
     }
 
     .su-alert__heading {
-      @include margin(0 null 0.3rem);
+      @include margin(0 null calculate-font-size(3px));
 
       @include grid-media('md') {
-        @include margin(0.3rem null null);
+        @include margin(calculate-font-size(3px) null null);
       }
     }
 
--- a/core/src/scss/utilities/functions/_fonts.scss
+++ b/core/src/scss/utilities/functions/_fonts.scss
@@ -1,0 +1,15 @@
+@charset "UTF-8";
+
+///
+/// Calculate font-size to use rems based upon 
+/// the root font size ($su-root-font-size) 
+/// set in core/src/scss/utilities/variables/core/_typography.scss
+///
+/// @name calculate-font-size
+///
+/// @param {String} $size-in-pixels - The value to be converted from pixels to rems.
+
+@function calculate-font-size($size-in-pixels) {
+  $font-size: $size-in-pixels / $su-root-font-size;
+  @return #{$font-size}rem;
+}
--- a/core/src/scss/utilities/functions/index.scss
+++ b/core/src/scss/utilities/functions/index.scss
@@ -8,6 +8,7 @@
 'breakpoint/index',
 'color',
 'flex/index',
+'fonts',
 'modular-scale/index',
 'string/index',
 'type-checks/index';


core/src/scss/utilities/variables/core/_typography.scss
```diff
@@ -12,7 +12,7 @@
 /// @name $su-root-font-size
 ///
 /// @group variable
-$su-root-font-size: 62.5%;
+$su-root-font-size: 10px;
```
sherakama: Thanks for submitting this PR but before you get any further I am concerned with this change. 

Older browsers don't scale fonts as well when they are set in PX values. The `62.5%` was chosen specifically because the default font size of the majority of browsers is 16px.  

( 62.5 / 100 ) * 16 = 10px

So the base font size is equivalent to 10px when set to 62.5% except when a user overrides their default browser settings to a font size other than 16px. When a user has a different default font size this percentage will scale the font size calculations across the board evenly so that all fonts scale up or down the same ratio. Setting the font size to a hard pixel value will disable the users' ability to scale their font sizes and hinder those with visual impairments. 

You should be able to use rems to get the correct size font wherever you need it. eg:

0.3 rem = 3px
1 rem = 10px
1.5 rem = 15px
2 rem = 20px

sherakama:
[Tugboat](https://tugboat.qa) has finished building the preview for this pull request!

Link:
- https://pr619-egoby3xprfopvyrcva2x2blqjypr1zvl.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/5e4459a4e74a0eb487b9c2be

Visual Diffs:
- https://dashboard.tugboat.qa/5e4459a4e74a0eb487b9c2be/visualdiffs

sherakama:
[Tugboat](https://tugboat.qa) has finished building the preview for this pull request!

Link:
- https://pr619-egoby3xprfopvyrcva2x2blqjypr1zvl.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/5e4459a4e74a0eb487b9c2be

Visual Diffs:
- https://dashboard.tugboat.qa/5e4459a4e74a0eb487b9c2be/visualdiffs

core/src/scss/utilities/variables/core/_typography.scss
silverli: @sherakama I think you might have posted this before we were able to get our PR description done. Will you read that over and let us know what you think?

silverli: we could switch this to 100% for our proof of concept in this PR if needed.

sherakama:
Ah, 

I see what you are trying to do. Yes, If you change the `$su-root-font-size` to 100% you will get the base font size as 16px and will still be accessible. Unfortunately, this does scale everything up drastically as you outlined above. We chose the `<html>` font size of 10px because that makes math easier for us as a base of 10 is easier than a scale of 16. We then use another variable to set the `<body>` font size to a larger size based on the root font size. Have a look at: https://github.com/SU-SWS/decanter/blob/master/core/src/scss/elements/typography/_typography.scss. So our base font size is actually 16px or larger but our calculations are base 10.

We complicate font sizes and calculations at different breakpoints by providing different `<body>` sizes at different breakpoints. We call this modular typography. Have a look at: https://github.com/SU-SWS/decanter/blob/master/core/src/scss/elements/typography/_typography.scss#L35

This mixin uses the sass map https://github.com/SU-SWS/decanter/blob/master/core/src/scss/utilities/variables/core/_modular-typography.scss#L19 to set the body size. In this case, the default settings are: 

xs - sm screens: 16px fonts
md - xl screens: 18px fonts
2xl screens: 19px fonts

So the base font size at each of those breakpoints is a curated value. That is why you will notice that all text gets larger as the screen gets larger. If you set all the breakpoints of the `$su-modular-typography-base` variable to be the same the fonts would be consistently sized across all screen sizes. 

For example, if I used 1rem with the default values, the font size output it would be:

16px on xs-sm screens
18px on md-xl screens
and
19px on 2xl screens


So, if you wanted to reset the html base font size to 16px to match your sass library and maintain the Decanter font sizes you will have to override two variables.

example:
```
$su-root-font-size: 100%;

$su-modular-typography-base: (
  'xs': 1rem,
  'sm': 1rem,
  'md': 1.125rem,
  'lg': 1.125rem,
  'xl': 1.125rem,
  '2xl': 1.187rem
);
```

We definitely need more documentation around the modular typography as it compounds with difficulty when we start to talk about heading scales and the `modular-scale` mixin from bourbon. But that can be a future topic.

Let me know if this helps.

silverli:
@sherakama Thank you so much for the very thorough response. 

I still had one question around the underlying issue of using specific rems in Decanter styles that we discussed in the PR description. Even if you are using modular typography to set the body font-size, if you set the font-size directly on some element, with something like the links in the main navigation to a font-size of 1.8rem like you do here: https://github.com/SU-SWS/decanter/blob/master/core/src/scss/utilities/placeholders/components/nav/_nav-item-link.scss#L14 that font-size will not scale with the modular type.

I think that 1.8rem is always based on whatever your base `html` font-size is. So when the Decanter team worked on this main navigation, it was equal to 18px. But if we reset the base `html` font-size in our styles, that number becomes 28.8px which is a lot larger looking on the page. 

Heather and I really want to use the Decanter styles you all have written for the main navigation, secondary nav, brand bar, etc. but the issue is that since some of those rems are hard coded in the Decanter styles, everything looks too large for us.

We understand and are happy to switch over our base font-size to 10px or 62.5%, and maybe that's best for now, but that's why we also proposed the `calculate-font-size` function in this PR as an alternative idea.

yvonnetangsu:
@silverli Thanks for your interest in using Decanter. 

Yes, we do have certain places where font sizes are hard coded using rem. These are usually elements we want very fine control over the font size at every breakpoint, e.g., the main nav and the global footer.

Changing Decanter's `$su-root-font-size` would mean having to change every value in Decanter that uses rem and also for projects that use the latest version of Decanter. It doesn't mean we wouldn't look into how we can make things more flexible in the future, but that might be for a future release.

If you're thinking about using some of the Decanter branded elements, perhaps give our modular typography system a try as well since those elements were designed to be proportional to  Decanter's base font sizes.

heatherdesigns:
@sherakama and @yvonnetangsu

Thank you for taking time to consider our pull request and for sharing information regarding `$su-root-font-size` and your modular  typography system. We truly appreciate your thoughtfulness in considering our request.

Moving forward we have decided to switch over our base font-size from 16px to 10px and use the `calculate-font-size` mixin on our side to make those changes and future-proof our system.

sherakama:
Thanks for this suggestion and conversation. 

This does call out an inconsistency in our project and how we are treating font-sizes in some of our components. If we do provide the option of allowing one to change the base font-size then it would be nice if all of the text was modified in the same way. I will add a note to our backlog to review this so that we are providing consistency in the product.

PR #678, opened 2020-04-29
D8CORE-1632: Add variables to quote template

sherakama:
# READY FOR REVIEW 

# Summary
- Adds new variables to quote template to allow for different media options

# Review By (Date)
- May 1st?

# Urgency
- Low

# Steps to Test

1. Review code
2. Review render in build url

- D8CORE-1632
- SOE Person type

# See Also


jenbreese:
@sherakama can you restart the tugboat? Or are we totally over the quota?

core/src/templates/components/quote/quote.twig
```diff
@@ -15,9 +16,12 @@
  */
 #}
 <figure {{ attributes }} class="su-quote {{ modifier_class }}">
-  {%- if src is not empty -%}
-  <img class="su-quote__img" src="{{ src }}" alt="{{ alt }}">
-  {%- endif -%}
+  {%- block block_quote_media -%}
```
yvonnetangsu, line 19: Actually could this use the nested media.twig template, like what we did for the Card component?

sherakama: Oh, that might be an option. Let me see about that refactor.

sherakama:
> Just that question about nested media.twig, and if not, do we need any stying for {{ quote_media }}, eg, float:left?

Naw, I'd like this to be an open field, and developers can bring their own styles if they add something wacky to it.

yvonnetangsu:
> > Just that question about nested media.twig, and if not, do we need any stying for {{ quote_media }}, eg, float:left?
> 
> Naw, I'd like this to be an open field, and developers can bring their own styles if they add something wacky to it.

@sherakama Please feel free to merge this in first so you can use the open {{ quote_media }} field. We can always decide whether we want the nested media.twig later.

sherakama:
> @sherakama Please feel free to merge this in first so you can use the open {{ quote_media }} field. We can always decide whether we want the nested media.twig later.

Thanks, It doesn't appear I will be getting the cycles to get to that work so merging this as is because it doesn't hurt.

Issue #680, opened 2020-05-12
Add aria-expanded="false" for main nav component parent links on load


yvonnetangsu:
**Is your feature request related to a problem? Please describe.**
This is a suggestion from Level Access. Currently in our main nav component, the parent links (ones that act as a dropdown trigger) do not have aria-expanded="false" on load which doesn't inform screen readers that there are child elements.  This attribute is added after they've been clicked on and do change state appropriately when interacted - which is nice, but we need them right from the start also.

**Describe the solution you'd like**
We just need to tweak the twig file to add aria-expanded="false" if link is a parent link in Main Nav.

yvonnetangsu:
Merged.

PR #681, opened 2020-05-12
Add aria-expanded="false" to main nav parent links in the twig template

yvonnetangsu:
# READY FOR REVIEW 

# Summary
- Had a consulting session with Level Access. They suggested adding `aria-expanded="false"` to parent links (the ones that you click to trigger a drop down menu to open) in our Main Nav on load. Currently, the `aria-expanded` attribute is added only after a parent link has been clicked on.

# Needed By (Date)
- N/A

# Urgency
- Not urgent but sooner the better

# Steps to Test
- Not sure what's the best way to test since tugboat is over quota. I did the same modification in UComm's Redwood WP template and this change works while keeping all current JS working as expected. Might want to make sure this doesn't break anything on the Drupal side (should be fairly low risk but just in case).

- Decanter
- Redwood, UComm sites

- #680


sherakama:
Thank you.

PR #700, opened 2020-07-29
Chipping away at Tailwind related changes/utility classes/new variables

yvonnetangsu:
# NOT READY

# Summary
- Chipping away at what needs to be done; also me learning how Tailwind works
- Disable some core plugins that will likely not be used (e.g., "skew") to reduce generated file size
- Add Decanter overrides to the core plugins (e.g., Stanford color palette)
- Add Tailwind plugins when makes sense
- Will be adding things in small or large batches

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

https://stanfordits.atlassian.net/browse/ADAPT-537?atlOrigin=eyJpIjoiMzExYTNkN2I2Mjk0NDM5M2IzMzczNWMwYmI1MDQ0MjAiLCJwIjoiaiJ9
- Other PRs
- #701 

# See Also
- @katrialesser


src/plugins/theme/boxShadow.js
```diff
@@ -7,7 +7,5 @@ module.exports = function () {
     'flat': 'none',
     sm: '0px 1px 2px rgba(0, 0, 0, 0.25)',
     md: '0px 10px 20px rgba(0, 0, 0, 0.15), 0px 6px 6px rgba(0, 0, 0, 0.2)',
-    lg: '0px 14px 28px rgba(0, 0, 0, 0.2), 0px 5px 10px rgba(0, 0, 0, 0.22)',
-    xl: '0px 19px 38px rgba(0, 0, 0, 0.25), 0px 5px 12px rgba(0, 0, 0, 0.18)',
```
sherakama, line 11: Why remove these?

src/plugins/theme/colors.js
```diff
@@ -1,45 +1,108 @@
 /**
- * Stanfords Default Color Pallet.
+ * Stanford's default color palette (updated with new Identity Guide release)
```
sherakama, line 2: Thanks for doing this.

src/plugins/theme/fontFamily.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Font Family Defaults.
+ */
+module.exports = function () {
+  return {
+    mono: ['Roboto Mono', 'Menlo', 'Courier New', 'monospace'],
+    sans: ['Source Sans Pro', 'Helvetica Neue', 'Helvetica', 'Arial', 'sans-serif'],
+    serif: ['Source Serif Pro', 'Georgia', 'Times', 'Times New Roman', 'serif'],
+    slab: ['Roboto Slab', 'Georgia', 'Times', 'Times New Roman', 'serif'],
```
sherakama, line 9: No more handwriting or Sanskrit? For backwards compatibility, we may need to add them back in.

src/plugins/theme/screens.js
```diff
@@ -3,13 +3,11 @@
  */
 module.exports = function () {
   return {
-    xs: '0px',
```
sherakama, line 6: Why no xs option?

src/plugins/theme/fontFamily.js
yvonnetangsu: Yeah this is the "ideal" version - without worrying about backward compatibility version we talked about :) Dana, Kerri and I think that we might be completely removing Kalam, and seems like the Sanskrit font doesn't belong in Decanter (more of a one-off thing).

src/plugins/theme/boxShadow.js
yvonnetangsu: We found that we never touched the two deepest shadows.

src/plugins/theme/screens.js
yvonnetangsu: In TW's default config, they don't explicitly include the 0 for the smallest breakpoint. I might change my mind later about some of these things though.

sherakama:
Thanks for the comments. Fast and loose at this point. Let's try it out.

yvonnetangsu:
> Thanks for the comments. Fast and loose at this point. Let's try it out.

Totally open to revert some of these changes if needed. I was mostly just trying things out :) Thanks for reviewing!

PR #705, opened 2020-08-11
V7 - Tailwind - Shea

sherakama:
# READY FOR REVIEW 

# Summary
- Try and see if you can get this set up.


yvonnetangsu:
I merged your branch into mine and combined some of our overlapping component stuff (brand bar, logo). I made some changes to centered container so it matches our current v6 behavior (added to some places in the sample html) - there is an extra breakpoint for that which is 1500px + 2*screen-margins-2xl which is 1700px. Please feel free to merge my branch into yours also - so we can merge yours into the v7 branch 😄  This is going pretty well - exciting!

PR #708, opened 2020-08-12
Chipping away at Tailwind stuff

yvonnetangsu:
# NOT READY

# Summary
- TL;DR - what's this PR for?

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


package.json
```diff
@@ -32,6 +32,7 @@
     "tailwindcss": "^1.6.2"
   },
   "devDependencies": {
+    "@fullhuman/postcss-purgecss": "^2.3.0",
```
sherakama, line 35: I don't think we can use this here as we won't be using all of our generated css.

src/css/index.css
```diff
@@ -1,15 +1,181 @@
...
+    @apply
+    inline-block
+    font-regular font-stanford
+    text-white;
+
+    line-height: 0.75;
+    -ms-font-feature-settings: "liga" 1;
+    font-feature-settings: "liga";
+    font-variant-ligatures: discretionary-ligatures;
+}
```
sherakama, line 45: Can these be moved into a component plugin?

Ideally, we would want everything to be in a plugin for maximum re-use and extensibility.

src/plugins/components/button/button.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function () {
```
sherakama, line 1: Niiiice.

src/plugins/components/button/button.js
```diff
@@ -0,0 +1,22 @@
...
+      '.button': {
+        padding: '1rem 2rem 1.15rem',
+        display: 'inline-block',
+        width: 'auto',
+        border: 'none',
+        backgroundColor: theme('colors.digital-red.default'),
+        color: theme('colors.white'),
+        textAlign: 'center',
+        fontFamily: sans,
+        fontWeight: '400',
```
sherakama, line 15:
```suggestion
        fontWeight: theme('fontWeight.normal'),
```

package.json
sherakama: We can use this in our projects that use this version of decanter but ootb we will need to have everything outputted.

src/css/index.css
yvonnetangsu: Agree - will do :)

package.json
yvonnetangsu: Yup - agree. Just testing. I've removed the purge stuff and saving the code for production use 😺

PR #773, opened 2020-11-24
D8CORE-2151: Remove empty html

cjwest:
# Not READY FOR REVIEW

# Summary
This removes unneeded HTML in the card when there's no content.
This removes the `<section>` tag that wasn't quite appropriate.

# Needed By (Date)
- 

# Urgency
Not urgent

# Steps to Test

1. check out this branch
2. Create a card with only the required content.
3. Verify that there's no additional HTML

D8Core

- JIRA ticket - [D8CORE-2151]( https://stanfordits.atlassian.net/browse/D8CORE-2151)
- Other PRs

# See Also

### DIFF
--- a/core/src/templates/components/card/card.twig
+++ b/core/src/templates/components/card/card.twig
@@ -62,6 +62,16 @@
   {%- endif -%}
 {%- endif -%}
 
+{# For variant where there's no card contents #}
+{% set card_contents_not_empty = false %}
+{% if card_super_headline is not empty or
+  card_headline is not empty or
+  card_body is not empty or
+  card_cta_label is not empty or
+  card_button_label is not empty %}
+  {% set card_contents_not_empty = true %}
+{% endif %}
+
 <article {{ attributes }} class="su-card {{ modifier_class }}">
   {# For variant where whole card is a link, add an <a> within the outer wrapper <div> #}
   {%- if allow_links == false %}
@@ -133,13 +143,14 @@
     {% endif %}
   {% endblock block_card_media -%}
 
-  <section class="su-card__contents">
+  {%- if card_contents_not_empty -%}
+  <div class="su-card__contents cjw">
     {# Card Super Headline #}
-    {%- block block_card_super_headline %}
-      {% if card_super_headline is not empty %}
+    {% if card_super_headline is not empty %}
+      {%- block block_card_super_headline %}
         <span>{{ card_super_headline }}</span>
-      {% endif -%}
-    {% endblock -%}
+      {% endblock -%}
+    {% endif -%}
 
     {# Card Headline #}
     {%- block block_card_headline  %}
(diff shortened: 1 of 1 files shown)


core/src/templates/components/card/card.twig
```diff
@@ -133,13 +143,14 @@
     {% endif %}
   {% endblock block_card_media -%}
 
-  <section class="su-card__contents">
+  {%- if card_contents_not_empty -%}
+  <div class="su-card__contents cjw">
```
pookmish, line 147:
```suggestion
  <div class="su-card__contents">
```

pookmish (changes requested):
The logic seems fine to me. Please assign to someone that maintains the decanter project to ensure its acceptable.

core/src/templates/components/card/card.twig
```diff
@@ -133,13 +143,14 @@
     {% endif %}
   {% endblock block_card_media -%}
 
-  <section class="su-card__contents">
+  {%- if card_contents_not_empty -%}
```
sherakama, line 146: Early on we asked if a card should ever be if it doesn't have a body. The general answer was "no" if the card does not have a body it is not a card. What is the use case for this pattern?

core/src/templates/components/card/card.twig
```diff
@@ -133,13 +143,14 @@
...
+  <div class="su-card__contents">
     {# Card Super Headline #}
-    {%- block block_card_super_headline %}
-      {% if card_super_headline is not empty %}
+    {% if card_super_headline is not empty %}
+      {%- block block_card_super_headline %}
```
sherakama, line 150: Blocks should wrap the conditional content statement. Blocks are not conditional in twig templates and will always be replaced/rendered if overridden from a parent template.

core/src/templates/components/card/card.twig
```diff
@@ -153,13 +164,16 @@
...
+        
       {% endif -%}
     {% endblock -%}
 
-    {%- block block_card_body %}
-      {# An open field for body content #}
-      {{ card_body }}
-    {% endblock -%}
+    {# An open field for body content #}
+    {% if card_body is not empty %}
```
sherakama, line 172: If statement is not needed here as card_body won't render anything if it is empty anyhow.

sherakama (changes requested):
Feedback left. Thanks for working on this.

core/src/templates/components/card/card.twig
sherakama: If left in this order, a parent template would have to assign set card_super_headline to something truthy AND override the block statement with their logic.

core/src/templates/components/card/card.twig
cjwest: @sherakama 
In https://stanfordits.atlassian.net/browse/D8CORE-2151, Ishi recommended removing the empty space. This seemed to be the most appropriate way of removing this space.

Please advise.

sherakama: Implement field validation on the card paragraph to require one of the values for the body area in cards and use the media paragraph for when you only want to display an image or video.

cjwest:
@sherakama,
Will you please take a look at this for another round of feedback?

sherakama (changes requested):
@cjwest I thought we agreed we wouldn't hide the card body content in this template and that the fix for the issue would be to use the appropriate pattern?

cjwest:
Hi @sherakama,
Thanks for the clarification. Looks like I mis-understood.

PR #775, opened 2020-11-24
D8CORE-2529: changing the condition that makes the link not appear

jenbreese:
# READY FOR REVIEW

# Summary
- This is a change to the way the link is made null. It should be added so this request can happen: https://github.com/SU-SWS/jumpstart_ui/pull/82. As it is currently, the link is always null, so I changed it to be only null if the media is null.  

# Needed By (Date)
- 12/2

# Urgency
- high

# Steps to Test

1. pull in this change. 
2. Verify the media works with image and link, image and no link. If there is a link and no images then the link doesn't appear.
3. I'm not sure how to test out this change with the other change https://github.com/SU-SWS/jumpstart_ui/pull/82 change
4. Need guidance here. :) 

- Decanter
- JumpstartUI and https://github.com/SU-SWS/jumpstart_ui/pull/82

- D8CORE-2529
- https://github.com/SU-SWS/jumpstart_ui/pull/82
- D8CORE-2105
- @sherakama @pookmish 

# See Also

### DIFF
--- a/core/src/templates/components/media/media.twig
+++ b/core/src/templates/components/media/media.twig
@@ -29,11 +29,6 @@
  */
 #}
 
-{# Set link to null if there is no content to prevent an empty link #}
-{%- if media_image_src is empty and media_audio_src is empty and media_video_src is empty -%}
-  {%- set media_link = false -%}
-{%- endif -%}
-
 <figure {{ attributes }} class="su-media {{ modifier_class }}">
   {# Option to make the media element a clickable link #}
   {%- if media_link %}
--- a/core/src/templates/components/media/media.twig
+++ b/core/src/templates/components/media/media.twig
@@ -30,7 +30,7 @@
 #}
 
 {# Set link to null if there is no content to prevent an empty link #}
-{%- if media_link is not empty and media_image_src is empty and media_audio_src is empty and media_video_src is empty -%}
+{%- if media_image_src is empty and media_audio_src is empty and media_video_src is empty -%}
   {%- set media_link = false -%}
 {%- endif -%}
 


sherakama (approved):
Looks good to me.

core/src/templates/components/media/media.twig
```diff
@@ -30,7 +30,7 @@
 #}
 
 {# Set link to null if there is no content to prevent an empty link #}
-{%- if media_link is not empty and media_image_src is empty and media_audio_src is empty and media_video_src is empty -%}
+{%- if media_image_src is empty and media_audio_src is empty and media_video_src is empty -%}
```
sherakama, line 33: The media link doesn't have to be tied to the media source. I would almost go the other way and remove all the src checks.

jenbreese: What I'd like to happen is that the link does not appear if there is no image. Right now, the link will appear even if it is the only bit of content provided. My thought was that I'd look to see if the media source was empty and make the link empty. Would it make sense instead to have all the fields as required?

sherakama: I'd rather have these components have less logic in them and use explicitly set variables. If this template gets passed a variable, then it should use it. 

So if you want to avoid having a link when there is no image, you would not pass in a media_link variable.

sherakama: Does Drupal or something else make that difficult?

jenbreese: I hadn't thought about it that way before. Does it make sense to you to have the media a required field?

jenbreese: If a client wanted a link or a caption without an image, they'd use a different paragraph type. If there's a link, there needs to be an image to link, right?

jenbreese: Ok. We talked more in the standup and I see the way. I'll revert what I did. :)

sherakama (approved):
I like this better. @yvonnetangsu thoughts?

yvonnetangsu:
> I like this better. @yvonnetangsu thoughts?

@sherakama @jenbreese Looks good to me!

yvonnetangsu (approved):
Looks good to me!

jenbreese:
Thank you both!

sherakama (approved):

PR #795, opened 2021-02-04
ADAPT 1756 - Continuous v7 prep work

yvonnetangsu:
# NOT READY

# Summary
- TL;DR - what's this PR for?

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


yvonnetangsu:
Screen Shot 2021-02-03 at 10 26 46 PM
👀  I did npm run publish - should I try something else instead? @sherakama

sherakama:
I think we may have been using `npm run build` so that the css is readable and usable until we get documentation about all the classes.

sherakama:
These look good. I am happy to keep them rolling in if you're ready.

yvonnetangsu:
> These look good. I am happy to keep them rolling in if you're ready.

@sherakama I saw your comment late so I pushed some more stuff before the dev chat. I actually removed the whole "button" folder in components in one commit, since we have html `<button>` styles in the base css so those are taken care of if people want their buttons to look somewhat decent. What do you think?

sherakama (approved):
GTG

sherakama:
If we bring the su-button back later we can. For those that want to use just the utility classes and not the base styles.

PR #800, opened 2021-02-11
ADAPT-000: Link color utility

sherakama:
# READY FOR REVIEW

# Summary
- For Discussion
- This is to tackle changing backgrounds, text-color, and link color on the wrapper container.

# Needed By (Date)
- Whenever
- 
# Urgency
- None

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


yvonnetangsu (approved):
Interesting.... I think this would be useful as long as we use it carefully since it would override a utility class directly place on a link that's within that parent with the su-link-xx class. Nice tool for those WYSIWYG for sure though. GTG - we can always tweak later if we need to.
Screen Shot 2021-02-11 at 9 42 25 AM

sherakama:
Yeah, I am not super fond of this approach but lacking the ability to get access to the individual links in something like a wysiwyg this might be a good enough get out of jail free card.

sherakama:
After a little more research this morning I am not finding anything to influence me in a different direction at this point. Merge and use with caution for now.

yvonnetangsu:
Sounds good to me - thanks for adding this!

PR #808, opened 2021-03-02
Adapt-1856: Misc v7 changes take 2

yvonnetangsu:
# READY FOR REVIEW

# Summary
- More misc additions to v7 - line height, border-radius, font-size, borderColor variants
- Add `su-basefont-21` for cases where e.g., a non-21px base font size is set on the `<body>` element but some section within the site need a `su-basefont-21` override.
- Add `last` and `first` variants for margin since we often set different margins for `first-child` and `last-child`, e.g., no margin-bottom on the last `<li>` in a `<ul>`
https://tailwindcss.com/docs/hover-focus-and-other-states#last-child

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Look at code
2. Check that the preview build is not broken

- JIRA ticket
- Other PRs

# See Also

### DIFF
--- a/src/plugins/base/base.js
+++ b/src/plugins/base/base.js
@@ -11,16 +11,6 @@ module.exports = function () {
     // Heading defaults.
     var headings = config('theme.decanter.typography.headings');
 
-    // Input base styles.
-    var inputBase = {
-      display: 'block',
-      maxWidth: config('theme.decanter.base.inputMaxWidth'),
-      borderColor: config('theme.colors.foggy.dark'),
-      color: config('theme.colors.black.DEFAULT'),
-      fontSize: config('theme.decanter.base.inputFontSize'),
-      lineHeight: config('theme.lineHeight.snug'),
-    };
-
     // Add the elements.
     addBase({
       'html': {
--- a/src/plugins/theme/borderRadius.js
+++ b/src/plugins/theme/borderRadius.js
@@ -4,7 +4,8 @@
 module.exports = function () {
   return {
     'none': '0',
-    DEFAULT: '0.3rem', // form input border-radius
+    'input': '0.3rem', // form input border-radius
+    '7': '0.7rem', // Local Footer login button radius
     full: '9999px', // TW convention for circular
   }
 }
--- a/src/plugins/theme/borderRadius.js
+++ b/src/plugins/theme/borderRadius.js
@@ -4,7 +4,7 @@
 module.exports = function () {
   return {
     'none': '0',
-    'input': '0.3rem', // form input border-radius
+    DEFAULT: '0.3rem', // form input border-radius
     '7': '0.7rem', // Local Footer login button radius
     full: '9999px', // TW convention for circular
   }
--- a/src/plugins/theme/fontSize.js
+++ b/src/plugins/theme/fontSize.js
@@ -18,8 +18,8 @@ module.exports = function () {
     '23': '2.3rem',
     '24': '2.4rem',
     '25': '2.5rem',
-    '09em': '0.9em',
     // Modular scale steps
+    '-m1': '0.9em',
     'm0': '1em',
     'm1': '1.25em',
     'm2': '1.56em',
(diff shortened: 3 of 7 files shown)


src/plugins/theme/borderRadius.js
```diff
@@ -4,7 +4,8 @@
 module.exports = function () {
   return {
     'none': '0',
-    DEFAULT: '0.3rem', // form input border-radius
```
sherakama, line 7: No more default? Why is this being removed?

src/plugins/base/base.js
```diff
@@ -11,16 +11,6 @@ module.exports = function () {
...
 
-    // Input base styles.
-    var inputBase = {
-      display: 'block',
-      maxWidth: config('theme.decanter.base.inputMaxWidth'),
-      borderColor: config('theme.colors.foggy.dark'),
-      color: config('theme.colors.black.DEFAULT'),
-      fontSize: config('theme.decanter.base.inputFontSize'),
-      lineHeight: config('theme.lineHeight.snug'),
-    };
```
sherakama, line 22: Was this not being used?

src/plugins/theme/fontSize.js
```diff
@@ -18,6 +18,7 @@ module.exports = function () {
     '23': '2.3rem',
     '24': '2.4rem',
     '25': '2.5rem',
+    '09em': '0.9em',
```
sherakama, line 21: This seems like a big outlier. Tell me more about this.

src/plugins/base/base.js
yvonnetangsu: I decided to add .su-input-base as a TW "component" (since we have the different html base font compared to TW default), so removing it from the base.

src/plugins/theme/borderRadius.js
yvonnetangsu: I'm on the fence about this. It seems the 0.3rem is only used for input fields, so I feel like calling it our default might be a little misleading, but I could be convinced otherwise 😄

sherakama: https://tailwindcss.com/docs/border-radius#border-radiuses

I still think we need to provide a `DEFAULT` option to tailwind. It is used in the `su-rounded` class. You can choose the sensible default that you like but I think we should still have one.

src/plugins/theme/fontSize.js
yvonnetangsu: From past experience, we use 0.9em quite often for those "slightly smaller than base" font styles. We have 1.4 to 2.5 rem here, then the smaller `em` step we have is `su-text-m0 = 1 em`. I kind of want a `em` one that's just a tad smaller than 1 em.

src/plugins/base/base.js
sherakama: SG2M

src/plugins/theme/borderRadius.js
yvonnetangsu: Sounds good - will have it =0.3rem then

src/plugins/theme/fontSize.js
sherakama: That makes sense but it breaks the naming conventions here. Could this be `-m1` for negative modular 1?

src/plugins/theme/borderRadius.js
yvonnetangsu: Thinking about this... I'm not completely sure about the 7px radius on the local footer login button. Will talk to @kerri-augenstein and see if I can convince her to maybe have this also equal to 0.3rem 😄  - but that's for later.

src/plugins/theme/fontSize.js
yvonnetangsu: I was debating about that since currently the m steps follow the 1.25 ratio, buuuuuut, -m1 is easier to remember and makes sense to not have too many outliers in the class names. So -m1 it is 😄

sherakama (approved):
GTG

PR #817, opened 2021-04-08
ADAPT-578: modified <a> text color for Info, Success and default Alerts

andrewbaraniuk:
# READY FOR REVIEW

# Summary
- Fixes <a> text color for the default Alert which currently looks like this - https://www.awesomescreenshot.com/image/7155716?key=68fb201780bf975cafa845ff50d471f8

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also

### DIFF
--- a/core/src/scss/components/alert/_alert.scss
+++ b/core/src/scss/components/alert/_alert.scss
@@ -58,15 +58,30 @@
     font-size: 1.7rem;
 
     .su-far,
-    .su-fas {
+    .su-fas,
+    .fas {
       @include margin(null null null 0.5em);
     }
 
     i,
     .su-far,
-    .su-fas {
+    .su-fas,
+    .fas {
       font-style: normal;
       text-decoration: none;
+    }
+
+    &.su-text-black {
+      &:hover {
+        background: $su-color-digital-blue-light;
+        color: $su-color-black;
+      }
+    }
+    &.su-text-white {
+      &:hover {
+        background: $su-color-white;
+        color: $su-color-black;
+      }
     }
 
   }
@@ -172,6 +187,21 @@
 // Optional open footer area.
 // -----------------------------------------------------------------------------
 .su-alert__footer {
+  &.su-link-black-true {
+    a {
+      &:hover {
+        background: $su-color-digital-blue-light;
+      }
+    }
+  }
+  &.su-link-black-false {
+    a {
+      &:hover {
+        background: $su-color-white;
+        color: $su-color-black;
+      }
+    }
+  }
 
   a {
     color: color(dark-text, $su-alert-colors);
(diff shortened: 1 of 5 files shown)


yvonnetangsu:
@andrewbaraniuk - Good catch! I think we never updated the link colors in Decanter 6. Thanks for this PR. Let me check with our designers on hover/focus states on these links (don't think we have it documented currently) and I'll get back to you.

core/src/scss/components/alert/_alert.scss
```diff
@@ -172,6 +187,21 @@
...
+    }
+  }
+  &.su-link-black-false {
+    a {
+      &:hover {
+        background: $su-color-white;
+        color: $su-color-black;
+      }
+    }
+  }
```
yvonnetangsu, line 204: We don't need to worry about the CTA link in the alert footer so please remove these changes.

core/src/scss/components/alert/_alert.scss
```diff
@@ -58,17 +58,32 @@
...
     .su-far,
-    .su-fas {
+    .su-fas,
+    .fas {
       font-style: normal;
       text-decoration: none;
     }
 
+    &.su-text-black {
+      &:hover {
```
yvonnetangsu, line 75: We need to apply these changes for the :focus state also.

core/src/scss/components/alert/_alert.scss
```diff
@@ -58,17 +58,32 @@
...
+        background: $su-color-digital-blue-light;
+        color: $su-color-black;
+      }
+    }
+    &.su-text-white {
+      &:hover {
```
yvonnetangsu, line 81: Same here, please add support for :focus state

yvonnetangsu: Also, we have a file `._alert--text-light.scss` in the same folder as _alert.scss, so it would probably make sense to move this part to that file. And you could use the .su-alert--text-light class on the Giving site alerts with dark background/light text. Thanks, @andrewbaraniuk !

yvonnetangsu (changes requested):
Just those few things in the comment - thank you, @andrewbaraniuk ! 😄

yvonnetangsu:
Forgot one - please add a `letter-spacing: 0.1em` to the alert label (eg, WARNING, SUCCESS) and the Dismiss button text as well. Thanks!

andrewbaraniuk:
> letter-spacing: 0.1em

Done!

yvonnetangsu (approved):
I pushed up some final changes. I think we're good to go. Thanks!

yvonnetangsu:
@sherakama Think we could punch a release for this one so we could use it for Giving?

sherakama:
Sure

sherakama:
@yvonnetangsu https://github.com/SU-SWS/decanter/releases/tag/6.2.3

yvonnetangsu:
@sherakama Thanks a bunch!

PR #834, opened 2021-05-26
Added 4,6,8 pixel border widths

sherakama:
# READY FOR REVIEW

# Summary
- Add a few more border width options to stay in alignment with core TW options
- Also, could use the 4px stroke for the 'fancy hover' style of link

# Needed By (Date)
- Whenever

# Urgency
- Low

# Steps to Test

1. Review code
2. Review preview build &/or config viewer


yvonnetangsu (approved):
GTG. Fancy hover was created with pseudo element though so not actually a border (can't animate border length span) 😄  I don't mind having more border widths though.

PR #851, opened 2021-09-17
ADAPT-000: Remove link-underline utility class

sherakama:
# READY FOR REVIEW

# Summary
- Removes link-underline utility classes as they don't work for JIT mode
- Updated to node version 16 and updated a few packages
 
# Steps to Test

1. Review code changes
2. Review preview build
3. Review upgrade.md dox

- Decanter React
- Alumni

- JIRA ticket
- Other PRs

# See Also


package.json
```diff
@@ -1,10 +1,19 @@
...
-  "license": "MIT",
+  "license": "GPL-3.0-or-later",
+  "keywords": [
+    "CSS",
+    "Decanter",
+    "Stanford",
+    "TailWindCSS"
+  ],
+  "source": "src/decanter.js",
+  "main": "dist/decanter.js",
```
yvonnetangsu, line 15: Just a question about this - do you mean decanter.css here since we don't have a decanter.js file in dist/?

yvonnetangsu (commented):
Just that one question - the rest looks good to me!

sherakama:
Aborting as there is no good way to make this plugin work with JIT mode. We are going to have to remove it.

yvonnetangsu:
> Aborting as there is no good way to make this plugin work with JIT mode. We are going to have to remove it.

Aww...bummer. Thank you for testing though! We'll find ways to work around that.

sherakama:
@rebeccahongsf @yvonnetangsu re-ready for review.

package.json
sherakama: Yeah, that!

yvonnetangsu (approved):
Thanks for investigating, @sherakama . GTG - bummer this doesn't work with JIT.
I suspect we also need to remove the link-color and link-fontweight utilities, but we could address that in another PR perhaps?
https://github.com/SU-SWS/decanter/blob/v7/src/plugins/utilities/link/link.js
https://github.com/SU-SWS/decanter/blob/v7/src/plugins/utilities/link/link-fontweight.js

sherakama:
Font weight should be fine but the hovers in link will have to go...

sherakama:
> Thanks for investigating, @sherakama . GTG - bummer this doesn't work with JIT.
> I suspect we also need to remove the link-color and link-fontweight utilities, but we could address that in another PR perhaps?
> [`v7`/src/plugins/utilities/link/link.js](https://github.com/SU-SWS/decanter/blob/v7/src/plugins/utilities/link/link.js)
> [`v7`/src/plugins/utilities/link/link-fontweight.js](https://github.com/SU-SWS/decanter/blob/v7/src/plugins/utilities/link/link-fontweight.js)

Good catch. The font-weight should be fine as it doesn't have the hacky variant classes (hover, focus) but the others have had their hacks removed. I restored the main selector for now but these should really be considered deprecated at this point. We will remove them from our other work first before completely removing them from this package.

yvonnetangsu:
@sherakama This looks good to me. Thank you for the fix!

PR #856, opened 2021-10-14
Release 6.2.5 Updates.

sherakama:
# READY FOR REVIEW

Can you review this `npm run publish` output?


yvonnetangsu:
@sherakama Sure, I'll review after standup. Thanks for this!

sherakama:
Not sure why they were marked as removed as they do appear to be there.
https://github.com/SU-SWS/decanter/blob/r625/core/dist/assets/fa-brands-400.eot

PR #864, opened 2022-02-28
ADAPT-4155: Update to tailwind v3

Lyafel:
Not ready for review

# Summary
- Updating decanter to tailwind v3

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


Lyafel:
@yvonnetangsu I looked some more into it and it does look like the css is there for test purposes only, alumni is pulling the js and compiles it's own css. So with that, looks like I could just redirect the tailwind config to look for the test html, and still have the css there only for demo.

src/css/index.css
```diff
@@ -28,4 +28,4 @@
  * If omitted, Tailwind will append these classes to the very end of
  * your stylesheet by default.
  */
-@tailwind screens;
```
sherakama, line 31: This file is mostly a placeholder/example but it is possible to import it as an expedient for projects that aren't doing it themselves, which should be rare.

tailwind.config.js
```diff
@@ -7,15 +7,15 @@
...
 
-var dir = path.resolve(__dirname, "src/plugins");
+const dir = path.resolve(__dirname, "src/plugins");
 
 module.exports = {
   // Our own prefix.
   prefix: "su-",
-  // Enable dark mode utility classes
-  darkMode: "media",
+  content: ["./dev/src/**/*.{html,js}"],
```
sherakama, line 18: I would like to avoid setting this in the base tw.config.js and use this file as a tailwind preset.

In the dev folder, for the tw.config.js file there, you should import this file as the preset and set the content path only there.

PR #868, opened 2022-04-06
DEC-000: Use NVM on GH Action.

sherakama:
# READY FOR REVIEW

### DIFF
--- a/.github/workflows/autorelease.yml
+++ b/.github/workflows/autorelease.yml
@@ -39,6 +39,10 @@ jobs:
           ref: ${{ github.base_ref }}
           fetch-depth: 0
 
+      - name: Set the correct Node version using nvm
+        shell: bash
+        run: nvm install
+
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease.yml
```diff
@@ -39,6 +39,11 @@ jobs:
           ref: ${{ github.base_ref }}
           fetch-depth: 0
 
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
+
```
sherakama, line 46: From: 
https://github.com/actions/setup-node/issues/32

sherakama:
This might help 🤷  but we also might have to update our dependencies.

.github/workflows/autorelease.yml
jenbreese: It looks right. I'm not sure how I could test it locally.

jenbreese (approved):
gtg

PR #871, opened 2022-04-08
ADAPT-4155: Refactor for TW3 and Presets

sherakama:
# READY FOR REVIEW

# Summary
- Preview of integration: https://github.com/SU-SWS/decanter-web/pull/96

# Needed By (Date)
- * Shrugs *

# Urgency
- None

# Steps to Test

1. Read code
2. Try out local environment
3. Review integration PR
4. Make suggestions or demands
5. Win

- Everything using V7
- This will be marked as an `rc` build as it is changing a fair bit


yvonnetangsu:
Screen Shot 2022-04-11 at 3 36 17 PM
Is that a new record? 😄 

I'll look at this tomorrow after finish another one of my My Account tickets 😄

sherakama:
Lol,

"Took a little off the top"

No rush on this. Feel free to throw whatever changes to the tailwind config you want into the mix.

yvonnetangsu:
@sherakama I'm thinking that there are perhaps 2 ways we could deal with root font sizes

1. We handle this with TW presets. One preset with root font size = 62.5% like we have in `base`, and another with root font size = 100% in `base`. There are only a handful of theme files that uses rem values, so we can create 2 sets of those to use with the 2 presets.
2. We use a CSS variable to set root font size for the project, so eg, with our current setting
```
:root {
  --root-font-size-factor: 0.625;
}
@tailwind base;
@tailwind components;
@tailwind utilities;
```
We then set HTML font size in base
```
html: {
        fontSize: 'calc(var(--root-font-size-factor) * 100%)',
},
```
Finally we use calculations for font sizes and spacings that uses rem like:
su-text-20 equals to  `calc(20rem / (var(--root-font-size-factor) * 16))`

This allow us to set the root font size back to 16px by using a --root-font-size-factor: 1, and everything else will be the same because it's calculated based on this variable (I tested this and it seems to work). Also, this way we don't have to go back and refactor the other projects.

sherakama:
@yvonnetangsu,

There are a couple of things that would benefit from the new preset approach that I would like to get an RC out for so please have a review of this and we can open up a new PR for the font size stuff.

yvonnetangsu:
> @yvonnetangsu,
> 
> There are a couple of things that would benefit from the new preset approach that I would like to get an RC out for so please have a review of this and we can open up a new PR for the font size stuff.

Yeah that sounds like a good plan. I'll look at this today 👍🏼

yvonnetangsu (approved):
Thank you for putting this together, @sherakama! Just pushed up a tiny change - please take a look, and then I believe it's GTG!! 🚀

PR #876, opened 2022-05-11
Add first and last variants

sherakama:
# READY

# Summary
- Adds first/last variants out of the box

Re: https://github.com/SU-SWS/saa_alumni/pull/329

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -44,6 +44,8 @@ module.exports = {
     plugin(({ addVariant }) => {
       addVariant('hocus', ['&:hover', '&:focus']);
       addVariant('children', '& > *');
+      addVariant('first', '&:first');
+      addVariant('last', '&:last');
     }),
 
     // 3rd Party Plugins;


yvonnetangsu:
Oh TW actually supports first/last variants out of the box, but when using with the custom `children` variant, I found that the order matters, e.g., I recently found that `first:children:` works, but `children:first` didn't (at least for my specific case in My Account). So that might be what we need to do 😄

sherakama:
> Oh TW actually supports first/last variants out of the box, but when using with the custom `children` variant, I found that the order matters, e.g., I recently found that `first:children:` works, but `children:first` didn't (at least for my specific case in My Account). So that might be what we need to do 😄

Huh, interesting.

yvonnetangsu:
> > Oh TW actually supports first/last variants out of the box, but when using with the custom `children` variant, I found that the order matters, e.g., I recently found that `first:children:` works, but `children:first` didn't (at least for my specific case in My Account). So that might be what we need to do 😄
> 
> Huh, interesting.

I think they're going to iron that kink out probably sooner or later. I'm hopeful 🤞🏼

sherakama:
Found this:
https://tailwindcss.com/docs/hover-focus-and-other-states#ordering-stacked-modifiers

Looks like order will matter for children

yvonnetangsu:
> Found this: https://tailwindcss.com/docs/hover-focus-and-other-states#ordering-stacked-modifiers
> 
> Looks like order will matter for children

Interesting....that makes sense!

sherakama:
Looks like a newer plugin for v3 also requires that the psuedo classes go before the children selectors: https://github.com/SamGoody/tailwind-children#psuedo-classes-psuedo-elements-and-state--eg-hover-after-open

yvonnetangsu:
> Looks like a newer plugin for v3 also requires that the psuedo classes go before the children selectors: https://github.com/SamGoody/tailwind-children#psuedo-classes-psuedo-elements-and-state--eg-hover-after-open

This might be more promising!

PR #905, opened 2023-02-24
Update TailwindCSS and other plugins; add utilities; update nvmrc

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Update nvmrc to use node v18
- Update TailwindCSS plugin and dependencies
- Add a few utilities

# Needed By (Date)
- Soon so I can install this version on new project

# Steps to Test

1. Look at the preview and check that it still works
1. Look at code


.nvmrc
```diff
@@ -1 +1 @@
-16
+18
```
yvonnetangsu, line 1: Most of our sites that uses decanter v7 is on node 18 now

src/plugins/theme/fontSize.js
```diff
@@ -4,7 +4,8 @@
  */
 module.exports = function () {
   return {
-    // su-text-12 to su-text-25 correspond to font sizes from 12px to 25px (in rem units)
+    // su-text-11 to su-text-30 correspond to font sizes from 12px to 30px (in rem units)
+    11: '1.1rem',
```
yvonnetangsu, line 8: We did go down to font-size: 11px on Directory

src/plugins/theme/fontSize.js
```diff
@@ -19,6 +20,11 @@ module.exports = function () {
     23: '2.3rem',
     24: '2.4rem',
     25: '2.5rem',
+    26: '2.6rem',
+    27: '2.7rem',
+    28: '2.8rem',
+    29: '2.9rem',
+    30: '3rem',
```
yvonnetangsu, line 27: The below 30px ones are useful for finetuning

tailwind.config.js
```diff
@@ -51,7 +51,14 @@ module.exports = {
         ':merge(.group):focus &',
         ':merge(.group):hover &',
       ]);
+      addVariant('group-hocus-visible', [
+        ':merge(.group):focus-visible &',
+        ':merge(.group):hover &',
+      ]);
```
yvonnetangsu, line 57: I'm fond of combining hover and focus-visible now

tailwind.config.js
```diff
@@ -51,7 +51,14 @@ module.exports = {
...
       addVariant('children', '& > *');
+      addVariant('children-hover', '& > *:hover');
+      addVariant('children-focus', '& > *:focus');
+      addVariant('children-focus-visible', '& > *:focus-visible');
```
yvonnetangsu, line 61: Add children interaction variants

.nvmrc
sherakama: If you want to enforce node 18 on the projects that are using this, you would add the engine version to `package.json`. I think it is fine not to require node 18 as I believe this package is still very much compatible with node 16. For us when working on this project, node 18 makes sense here.

sherakama (approved):
GTG

.nvmrc
yvonnetangsu: Got it - that sounds good. No enforcing it is 👍🏼 😄

PR #907, opened 2023-04-20
Provide sm and md gaps

pookmish:
Added sm and md gaps.

### DIFF
--- a/src/plugins/theme/gap.js
+++ b/src/plugins/theme/gap.js
@@ -3,7 +3,9 @@
  */
 module.exports = function () {
   return {
-    xs: '20px',
+    xs: '5px',
+    sm: '10px',
+    md: '20px',
     lg: '36px',
     xl: '40px',
     '2xl': '48px',


yvonnetangsu (changes requested):
@pookmish The reason that we have 20px for `xs` is that in the design system, the grid gap at XS breakpoint is 20px (same for SM and MD - which is why we omitted those).
I certainly don't mind us adding extra gap values, but changing `su-gap-xs` from 20px to 5px is going to break some of our sites 😄 

What do you think about adding instead:
```
5: 5px,
10: 10px,
```
and we leave `xs` unchanged? We have some of our existing spacing values this way too, eg, the `su-mt-10` etc. That might be more intuitive than adding `'2xs',` '`3xs'` etc.

pookmish:
Yeah I wondered. But 20px is pretty big for `xs` IMO. And that doesn't leave much difference for `sm` or `md`. I am inclined to remove those gap sizing entirely and go with normal tailwind gaps to allow a lot more flexibility.
Is there a way to extend instead of replace the tailwind classes?

yvonnetangsu:
> Yeah I wondered. But 20px is pretty big for `xs` IMO. And that doesn't leave much difference for `sm` or `md`. I am inclined to remove those gap sizing entirely and go with normal tailwind gaps to allow a lot more flexibility. Is there a way to extend instead of replace the tailwind classes?

Yes, I'll be happy to change this to extend the TW gaps 👍🏼 Life has gotten better since TW isn't using the legacy mode anymore.

We've been mainly using the gaps for larger layouts (eg, a card grid) thus why it's 20px at XS 😄  We have the custom `su-grid-gap` class which is basically combining `su-gap-xs su-gap-lg su-gap-xl su-gap-2xl`

yvonnetangsu:
@pookmish To extend the TW gap offerings, in the `tailwind.config.js` file `theme` section, simply move the require line for the gap file to inside the `extend: {}`. LMK if that works 😄 

One heads up, since the Decanter root font size is set to `10px`, if you're using the TW gap values, and if they're in units of `rem`, the resulting `px` value would be different from the documentation on the TW site, but the conversion would be simple. Let's say on TW site, if it says a certain gap value is `1.5rem` - the `px` value for Decanter would be `15px`.

yvonnetangsu:
@pookmish I'm going to talk to Kerri about the gaps this afternoon  - and will get back to you 😄

yvonnetangsu:
@pookmish I opened a new PR #908 for you to look at. LMK what you think and if you want additional values.

yvonnetangsu:
v7.1.1 now supports all the gap options with the values in the updated `spacing` theme below, so I think we can close this PR.
https://github.com/SU-SWS/decanter/blob/v7/src/plugins/theme/spacing.js

PR #908, opened 2023-05-03
v7 - Update TW to 3.3.2 and add Gap options

yvonnetangsu:
# READY FOR REVIEW

# Summary
- More Gap options
- Update TW to 3.3.2 (also Autoprefixer and postcss to latest)
- Uninstall the line-clamp plugin (in the previous PR we only remove the require line in the tailwind config file)

# Steps to Test

1. Look at what's changed
2. See if the gap values provided are ok
3. Look at preview and that things are not broken.

### DIFF
--- a/src/plugins/theme/gap.js
+++ b/src/plugins/theme/gap.js
@@ -3,9 +3,24 @@
  */
 module.exports = function () {
   return {
+    // The following are tied to our breakpoints in src/plugins/theme/screens.js
     xs: '20px',
     lg: '36px',
     xl: '40px',
     '2xl': '48px',
+    // Additional commonly used gutters for convenience
+    1: '1px',
+    5: '5px',
+    10: '10px',
+    15: '15px',
+    20: '20px', // Alias for gap-xs
+    30: '30px',
+    40: '40px', // Alias for gap-xl
+    50: '50px',
+    60: '60px',
+    70: '70px',
+    80: '80px',
+    90: '90px',
+    100: '100px',
   };
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,23 +11,33 @@
       "dependencies": {
         "@tailwindcss/aspect-ratio": "^0.4.2",
         "@tailwindcss/forms": "^0.5.3",
-        "@tailwindcss/line-clamp": "^0.4.2",
-        "tailwindcss": "^3.3.1"
+        "tailwindcss": "^3.3.2"
       },
       "devDependencies": {
         "auto-changelog": "^2.2.1",
-        "autoprefixer": "^10.4.13",
+        "autoprefixer": "^10.4.14",
         "eslint": "^7.14",
         "eslint-config-airbnb": "^18.2.1",
         "eslint-config-prettier": "^8.3.0",
         "eslint-config-standard": "^16.0.2",
         "eslint-plugin-prettier": "^3.4.0",
         "light-server": "^2.9.1",
-        "postcss": "^8.4.21",
+        "postcss": "^8.4.23",
         "postcss-import": "^13.0.0",
         "postcss-loader": "^4.0.0",
         "postcss-nested": "^5.0.1",
         "prettier": "^2.2.1"
+      }
+    },
+    "node_modules/@alloc/quick-lru": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz",
+      "integrity": "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==",
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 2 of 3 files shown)


src/plugins/theme/gap.js
```diff
@@ -3,9 +3,24 @@
...
+    15: '15px',
+    20: '20px', // Alias for gap-xs
+    30: '30px',
+    40: '40px', // Alias for gap-xl
+    50: '50px',
+    60: '60px',
+    70: '70px',
+    80: '80px',
+    90: '90px',
+    100: '100px',
```
yvonnetangsu, line 24: @pookmish What do you think about this list of options. I added the 5px and 10px that you added in your PR, and a bunch of additional ones (that we did use on ADAPT - mostly larger gutters for layouts) 😄 

Instead of extending the TW, I'm keeping it consistent with our spacing values, eg, su-mt-10 = 10px margin top not some random values. So here if you use su-gap-5, you'll get a gap of 5px.

sherakama: I like this approach. It is a bit of a mixed bag of numbers and letters but it aligns with other approaches and it doesn't cause any issue with backward compatibility.

yvonnetangsu:
@pookmish What do you think about these gap options and this approach? 😄 If you approve, we can get an autorelease out. cc @sherakama

src/plugins/theme/gap.js
pookmish: This does seem better, but I'm always frustrated that when Decanter replaces tailwind classes. There's so much documentation & tools for tailwind... then things like this rip out parts and the developer doesn't know why. There's no way to tell what parts of tailwind have been overridden by Decanter.
This is an improvement, but I stand firmly in the belief that this should always extend and never replace tailwind. Honestly, I have been very tempted to remove decanter in my project for this reason.

yvonnetangsu: Another way of course is we could add a table of values like on the TW website on the Decanter website so there's convenient access to the documentation. We do write documentation in the Decanter TW theme files, but I understand that it's not as convenient as going to the TW website, but if we could provide that in the Decanter website, perhaps that would be a better experience for developers?

sherakama: > This does seem better, but I'm always frustrated that when Decanter replaces tailwind classes

🤔 This is good feedback. This stems from the wanting to keep developers and designers 'on the system'. It was intentionally done as it would raise the flag with the developer that they are going off of the design pattern/system. The developer still could go off the system with the `[]` escape hatch but I do see how this would be annoying.

The pattern of pulling out parts is an intentional practice. However, we thought there would be more consistency between our project implementations and expected developers to go back to designers to get them to 'stay on the system'. In the real world, and with our diversity of work, we have a new system for every project and within those projects, we have plenty of exceptions. 

My new stance on our cross-project tools is not to be 'on brand' or 'on the system' as those things are too different on each project to thread the needle with a set of configurations, is to stay as an accelerant for building whatever it is you want. So when items like this get in your way, that is a problem.  

> and the developer doesn't know why

I hope the above explains why. 

> There's no way to tell what parts of tailwind have been overridden by Decanter

That is not true as you can easily look at the tailwind.config.js file to see. However, I agree that you shouldn't have to do that. We are working on documentation and it is coming. 

> should always extend and never replace

This seems like a reasonable thing to do. I wonder what the cross section of the TW design system and the Decanter design system is? Can we do this? Have you tried? Can you POC it for us? I'm open to moving in this direction and releasing it as a 7.1.

> I have been very tempted to remove decanter in my project for this reason

Dude, don't get all cancel culture on us. You're a part of this team and we value your contributions. You've got skills to make this project better and we'd like you to continue to work with us.

yvonnetangsu: Yeah second what @sherakama said - we appreciate your feedback and want to make this better. Hope we can work together and find a way 😄

sherakama: > So every time something doesn't work, i have to either navigate through the dozens of files to find the possible decanter file that rips out the class from tailwind. Even just looking for the correct color makes that a ridiculous task of finding the decanter folder among the node_modules directory.

If you're open to feedback I would suggest that your approach is not optimal. You know where the project is on github and can look at it directly: https://github.com/SU-SWS/decanter/blob/v7/tailwind.config.js

It is not as fancy as the tailwind cheatsheets site but if you have forgotten about the documentation we do have I would ask you to review it again. Perhaps it can help you with your color search. 

* https://decanter-v7.netlify.app/

> If it's not easy, it's a bad tool. 
> If it's more difficult, I will not use it. It's as simple as that.
>  Is Decanter documentation, walkthrough, YouTube videos as good as those that provide information on Tailwind? No.

This is not useful or actionable feedback. You very much know the status of this project and that it is a work in progress. You know we haven't released a stable version because we don't have the documentation to support it outside of our group. 

> removing the ability to use a class from tailwind makes this it's own thing and not useful.

I understand and appreciate your opinion on removing vs extending and I am open to moving this project in that direction. Where I think we will still run into a poor DX with that approach is on the documentation again. Can you build on that request for me and your comments on the lack of good documentation for Decanter? I'd like to know what value you have had with using the Decanter extension thus far (what has worked) and what form of documentation you would like to see. How would you document the delta between core TW and Decanter? Are you suggesting that unless Decanter can provide something as easy/searchable as the cheatsheet site that it will not be a useful tool?

pookmish: > If you're open to feedback I would suggest that your approach is not optimal. You know where the project is on github and can look at it directly: https://github.com/SU-SWS/decanter/blob/v7/tailwind.config.js

So I'm expected to navigate through all the dozens of files on github instead of `node_modules`? Then find the particular classes and figure out if it's being replaced from Tailwind. And then figure out which classes I'm actually supposed to use?
That's basically no different than navigating in an IDE. If there's a documentation site, then do I look at two websites to get the total perspective? 

If there was a documentation for Decanter, I would expect it to fully replace the Tailwind documentation. So every part of Tailwind could be referred in one location.

I've tried to provide suggestion to change various parts, but even though it's not in a stable release, there seems to be no desire to change the current behavior of the tool. Even something as simple as adjusting and adding `sm` and `md` isn't possible.

The only part of Decanter that has actually been helpful is the colors. But then again, I'm not able to use the colors from Tailwind so then I have to do the `[]` overrides in various places that could otherwise use standard tailwind classes. Besides the various base styling of elements, the only other classes decanter specific classes I used are `list-unstyled`.. Maybe that's all :thinking: 

Besides that, there's nothing I've used that Decanter makes easier, only more difficult. I even had to replace the logo with an SVG because the font made things more difficult because of some space at the bottom of the letters.

yvonnetangsu:
I have another approach I want to try which will provide more gap options. I'm going to close this PR and open another one.

PR #911, opened 2023-05-20
Change centered-container to use max width and allow nesting of classes

pookmish:
# READY FOR REVIEW

# Summary
- Change the way the `.centered-container` works. The current solution only works at the outer most container element, often children of `body` or such. Otherwise you get gutters inside gutters inside gutters. This change allows the user to nest the classes and properly set the element to have appropriate gutters at all breakpoints.

ex: 
```
<div class="cc">
   <div class="cc">
    <div class="cc">
       this element would be 450px inset from the window on each side while on 2xl screens, and 75px on each side on mobile. The proposed change would make sure this text would only have 25px on mobile and 150px on 2xl.
    </div>
  </div>
</div>
```

### DIFF
--- a/src/plugins/components/layout/centered-container.js
+++ b/src/plugins/components/layout/centered-container.js
@@ -2,7 +2,7 @@
  * Centered container.
  */
 module.exports = function () {
-  return function ({ addUtilities, theme }) {
+  return function ({ addComponents, theme }) {
     // Find and set the padding based on the screen margins
     const screens = theme('screens');
     const maxWdiths = {};
@@ -42,6 +42,6 @@ module.exports = function () {
       },
     };
 
-    addUtilities(components);
+    addComponents(components);
   };
 };
--- a/src/plugins/components/layout/centered-container.js
+++ b/src/plugins/components/layout/centered-container.js
@@ -30,7 +30,7 @@ module.exports = function () {
     const key = `@media only screen and (min-width: ${largestScreen + (largestGutter * 2)}px)`
     const components = {
       // Center an element horizontally.
-      '.centered': {
+      '.centered-container, .cc': {
         maxWidth: 'calc(100vw - ' + (smallestGutter * 2) + 'px)',
         width: '100%',
         marginLeft: 'auto',
--- a/src/plugins/components/layout/centered-container.js
+++ b/src/plugins/components/layout/centered-container.js
@@ -2,36 +2,38 @@
  * Centered container.
  */
 module.exports = function () {
-  return function ({ addComponents, theme }) {
+  return function ({ addUtilities, theme }) {
     // Find and set the padding based on the screen margins
     const screens = theme('screens');
-    const padding = {};
+    const maxWdiths = {};
 
-    // Create padding for each screen size which equals to the screen margin setting.
+    // Create max widths for each screen size which equals to the screen margin
+    // setting.
     const keys = Object.keys(screens);
     keys.forEach((key) => {
-      padding[`@screen ${key}`] = {
-        paddingLeft: theme(`decanter.screenMargins.${key}`),
-        paddingRight: theme(`decanter.screenMargins.${key}`),
-      };
+      if (theme(`decanter.screenMargins.${key}`)) {
+        const gutterWidth = parseInt(theme(`decanter.screenMargins.${key}`));
+
+        maxWdiths[`@screen ${key}`] = {
+          maxWidth: 'calc(100vw - ' + (gutterWidth * 2) + 'px)',
+        };
+      }
     });
 
     const components = {
       // Center an element horizontally.
       '.centered-container, .cc': {
-        paddingLeft: theme('decanter.screenMargins.xs'),
-        paddingRight: theme('decanter.screenMargins.xs'),
+        maxWidth: 'calc(100vw - ' + (parseInt(theme(`decanter.screenMargins.xs`)) * 2) + 'px)',
+        width: '100%',
         marginLeft: 'auto',
         marginRight: 'auto',
-        ...padding,
-        // At 1700px (2xl breakpoint + twice the screen margins at 2xl), the max container width stays at 1500px.
+        ...maxWdiths,
         '@media only screen and (min-width: 1700px)': {
-          paddingLeft: 'calc((100% - 1500px)/2)',
-          paddingRight: 'calc((100% - 1500px)/2)',
+          maxWidth: '1500px'
         },
       },
     };
 
-    addComponents(components);
+    addUtilities(components);
   };
 };
--- a/src/plugins/components/layout/centered-container.js
+++ b/src/plugins/components/layout/centered-container.js
@@ -6,30 +6,38 @@ module.exports = function () {
     // Find and set the padding based on the screen margins
     const screens = theme('screens');
     const maxWdiths = {};
+    let largestScreen = 0;
+    let largestGutter = 0;
+    let smallestGutter = 999999;
 
     // Create max widths for each screen size which equals to the screen margin
     // setting.
     const keys = Object.keys(screens);
     keys.forEach((key) => {
+      if (largestScreen < parseInt(screens[key])) largestScreen = parseInt(screens[key]);
+
       if (theme(`decanter.screenMargins.${key}`)) {
         const gutterWidth = parseInt(theme(`decanter.screenMargins.${key}`));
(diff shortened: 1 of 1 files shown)


src/plugins/components/layout/centered-container.js
```diff
@@ -2,36 +2,46 @@
  * Centered container.
  */
 module.exports = function () {
-  return function ({ addComponents, theme }) {
+  return function ({ addUtilities, theme }) {
```
sherakama, line 5: Why the change to Utility here?

src/plugins/components/layout/centered-container.js
```diff
@@ -2,36 +2,46 @@
...
     // Find and set the padding based on the screen margins
     const screens = theme('screens');
-    const padding = {};
+    const maxWdiths = {};
+    let largestScreen = 0;
+    let largestGutter = 0;
+    let smallestGutter = 999999;
```
sherakama, line 11: Just shy of a million?

src/plugins/components/layout/centered-container.js
```diff
@@ -2,36 +2,46 @@
...
+      }
+    });
+      
+    const key = `@media only screen and (min-width: ${largestScreen + (largestGutter * 2)}px)`
     const components = {
       // Center an element horizontally.
-      '.centered-container, .cc': {
-        paddingLeft: theme('decanter.screenMargins.xs'),
-        paddingRight: theme('decanter.screenMargins.xs'),
+      '.centered': {
```
sherakama, line 33: Why the change in name?

sherakama:
I like this change in behavior as it matches my expectation of how it should behave and how other CSS libs behave. A couple of questions in here first.

src/plugins/components/layout/centered-container.js
pookmish: > Use the addUtilities function to register simple static utilities that don’t support user-provided values

My belief is this matches the criteria. But I honestly think they accomplish the same things. :shrug:

src/plugins/components/layout/centered-container.js
pookmish: dr evil

src/plugins/components/layout/centered-container.js
```diff
@@ -2,36 +2,46 @@
...
-        paddingLeft: theme(`decanter.screenMargins.${key}`),
-        paddingRight: theme(`decanter.screenMargins.${key}`),
-      };
-    });
+      if (largestScreen < parseInt(screens[key])) largestScreen = parseInt(screens[key]);
+
+      if (theme(`decanter.screenMargins.${key}`)) {
+        const gutterWidth = parseInt(theme(`decanter.screenMargins.${key}`));
+        if (largestGutter < gutterWidth) largestGutter = gutterWidth;
+        if (smallestGutter > gutterWidth) smallestGutter = gutterWidth;
```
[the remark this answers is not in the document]
pookmish: well we don't know which one is the smallest or in which order they are coming in. I think its best to allow for as dynamic as it can be.

src/plugins/components/layout/centered-container.js
pookmish: Woops... that's a copy pasta error.

src/plugins/components/layout/centered-container.js
sherakama: https://tailwindcss.com/docs/adding-custom-styles

```
The base layer is for things like reset rules or default styles applied to plain HTML elements.
The components layer is for class-based styles that you want to be able to override with utilities.
The utilities layer is for small, single-purpose classes that should always take precedence over any other styles.
```

I believe we chose the components option for this one so that other utility classes would win if the developer chose to add different behaviours later.

pookmish: ahh. interesting! yeah i agree. I didn't realize that it would add them at different layer

pookmish:
I did this implementation and it's so much easier to just set the class on the drupal paragraph and I never have to worry about the container creating padding or what it's size or left/right side contains.

yvonnetangsu:
@pookmish @sherakama Was in meeting this morning so I need to catch up with the full thread of conversations, but, we did have a reason that we used the padding method for the existing `su-cc` class - it's a long story due to having it exactly match the Decanter v6 `centered-container` mixin and it's because of UComm's request for that approach.

Basically, using existing `su-cc`, it would allow one to do this
```
<div class="su-cc su-bg-digital-red">
Some text
</div>
```

It will create the max width container around the text, but still have the background color shown as edge to edge. UComm wanted to use as few nested container as possible so that's why we did it that way (instead of using outer container with bg color, and inner container with su-cc).

So, because of that history, directly modify the existing `su-cc` class would break backward compatibility for our sites 😬 . Are you open to adding your approach as `su-cc-alt` perhaps @pookmish?

yvonnetangsu:
@pookmish Personally I'd use your approach too if it's not because of backward compatibility and the UComm request back then.
cc @sherakama

pookmish:
> So, because of that history, directly modify the existing `su-cc` class would break backward compatibility for our sites grimacing . Are you open to adding your approach as `su-cc-alt` perhaps @pookmish?

Yet again, poor implementation for an obscure use with a single site due to legacy issues. Feel free to do what you wish to my PR. I'll just keep overridding more and more or using less of Decanter in my projects.

yvonnetangsu:
> > So, because of that history, directly modify the existing `su-cc` class would break backward compatibility for our sites grimacing . Are you open to adding your approach as `su-cc-alt` perhaps @pookmish?
> 
> Yet again, poor implementation for an obscure use with a single site due to legacy issues. Feel free to do what you wish to my PR. I'll just keep overridding more and more or using less of Decanter in my projects.

It's not a single site. We have half a dozen live sites using decanter v7 and making this change would require us to test all those sites and make changes if needed.

sherakama:
> Yet again, poor implementation for an obscure use with a single site due to legacy issues. Feel free to do what you wish to my PR. I'll just keep overridding more and more or using less of Decanter in my projects.

@pookmish here's your tiny violin. 



@yvonnetangsu, This sounds like we have two different use cases here and we have a need for a centered banner and a centered chromeless. And to add both, we would have a naming issue. the centered container should be the chromeless option IMHO but we should support the banner too.

Can we do both and come up with class names that make sense? I'm ok for a find-all-and-replace type change.

pookmish:
Ok, I won't be making contributions any longer. There's too much difficulty to make improvements.

sherakama:
@pookmish 

Thank you for meeting with me and talking this through. I apologize again for the inappropriate giphy but I am hoping you can write out a few answers to some of the questions we talked about and a couple of new ones to document your challenges. 

* What would you have liked the outcome of this PR to be if straight acceptance was not possible or if a disagreement on the legitimacy of existing use cases?
* How does the suggested compromise in this thread add technical debt? How does it negatively affect your work?
* How does "overridding more and more or using less of Decanter" negatively affect your work?
* What are your tradeoffs to abandoning Decanter? ie: what do you stand to gain/lose?

You also mentioned that overriding individual class names isn't possible and that it creates a merge of styles. I've flagged that as something to look into. 

thanks,

jbickar:
> @pookmish
> 
> Thank you for meeting with me and talking this through. I apologize again for the inappropriate giphy but I am hoping you can write out a few answers to some of the questions we talked about and a couple of new ones to document your challenges.
> 
>     * What would you have liked the outcome of this PR to be if straight acceptance was not possible or if a disagreement on the legitimacy of existing use cases?
> 
>     * How does the suggested compromise in this thread add technical debt? How does it negatively affect your work?
> 
>     * How does "overridding more and more or using less of Decanter" negatively affect your work?
> 
>     * What are your tradeoffs to abandoning Decanter? ie: what do you stand to gain/lose?
> 
> 
> You also mentioned that overriding individual class names isn't possible and that it creates a merge of styles. I've flagged that as something to look into.
> 
> thanks,

@pookmish, please respond in this thread.

sherakama:
Hey @pookmish,

We'd like to wrap this up with you and need your response. Please provide your response to the final questions and we can close this out and move along.

sherakama:
@pookmish It's been more than two weeks and I haven't heard back from you. Please wrap this up by answering the remaining questions. Thank you.

sherakama:
@pookmish this is a reminder about the thread you agreed to answer when we met in our 1:1 on Tuesday. Please answer the questions in https://github.com/SU-SWS/decanter/pull/911#issuecomment-1561620893

PR #918, opened 2023-07-11
v7 - Remove su- prefix; update sans and serif fonts; add variant group-hocus-within:

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Remove the `su-` prefix due to popular demand
- Update TW theme fontFamily to use Source Sans 3 and Source Serif 4. Leaving Source Sans Pro and Source Serif Pro as 2nd in lines for fallback for now to support incremental update of projects
- Add `group-hocus-within`: variant (`group-hover:` + `group-focus-within:`)
- TW version and other dependencies update
- Update preview site and docs to remove `su-` prefix

# Steps to Test

1. Look at the preview and see that it functions and looks the same; inspect to see that the TW classes now has no su- prefix
2. Look at code
3. Look at things called out

- This is related to the UComm announcement about the font change

### DIFF
--- a/fonts.css
+++ b/fonts.css
@@ -1,8 +1,7 @@
 @charset "UTF-8";
 
 /** Source Sans Pro, Source Serif Pro, Roboto Slab, Roboto Mono (regular only) */
-@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&family=Source+Serif+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap');
-
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+3:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&family=Source+Serif+4:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap');
 @font-face {
   font-family: "Stanford";
   src: url("https://www-media.stanford.edu/assets/fonts/stanford.woff") format("woff"), url("https://www-media.stanford.edu/assets/fonts/stanford.ttf") format("truetype");
--- a/fonts.css
+++ b/fonts.css
@@ -1,6 +1,6 @@
 @charset "UTF-8";
 
-/** Source Sans Pro, Source Serif Pro, Roboto Slab, Roboto Mono (regular only) */
+/** Source Sans 3, Source Serif 4, Roboto Slab, Roboto Mono (regular only) */
 @import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+3:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&family=Source+Serif+4:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap');
 @font-face {
   font-family: "Stanford";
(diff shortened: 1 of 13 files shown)


fonts.css
```diff
@@ -1,8 +1,7 @@
 @charset "UTF-8";
 
 /** Source Sans Pro, Source Serif Pro, Roboto Slab, Roboto Mono (regular only) */
-@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&family=Source+Serif+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap');
-
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+3:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&family=Source+Serif+4:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap');
```
yvonnetangsu, line 4: These font imports are only used for the preview site here; I replaced our sans and serif fonts, plus got rid of the 300 font weight since we never use them in the preview and won't (we don't even use that for our projects even).

src/plugins/theme/fontFamily.js
```diff
@@ -5,13 +5,15 @@ module.exports = function () {
   return {
     mono: ['"Roboto Mono"', 'Menlo', '"Courier New"', 'monospace'],
     sans: [
+      '"Source Sans 3"',
```
yvonnetangsu, line 8: Adding the SS3, but leaving the SS Pro as 2nd place for now

src/plugins/theme/fontFamily.js
```diff
@@ -5,13 +5,15 @@ module.exports = function () {
...
       '"Source Sans Pro"',
       '"Helvetica Neue"',
       'Helvetica',
       'Arial',
       'sans-serif',
     ],
     serif: [
+      '"Source Serif 4"',
```
yvonnetangsu, line 16: Same for the serif font - add the new, but leave the old one as 2nd place

tailwind.config.js
```diff
@@ -56,6 +53,10 @@ module.exports = {
         ':merge(.group):focus-visible &',
         ':merge(.group):hover &',
       ]);
+      addVariant('group-hocus-within', [
+        ':merge(.group):focus-within &',
+        ':merge(.group):hover &',
+      ]);
```
yvonnetangsu, line 59: Tested this out in Campaign and added for convenience

src/plugins/theme/fontFamily.js
sherakama: Why?

sherakama: Also, 
Don't we need to update the import as well?

https://github.com/SU-SWS/decanter/blob/v7/src/css/index.css#L4

yvonnetangsu: Just in case someone installs this latest version of Decanter but missing the note about the font change in the projects. They have to also update the font import in their own projects, or else it will render the other fall back font like Arial 😬 

It might not be obvious when they preview locally because the serif and sans fonts are not super different from the default fonts (times new romans or arial etc)

Open to suggestions though 😄

fonts.css
sherakama: https://github.com/SU-SWS/saa_alumni/blob/dev/src/styles/global.css#L1

We totally can and do use them in projects

yvonnetangsu: @sherakama You're right we need to update the font import in the index.css file. I only updated it here in this font.css file 😬 Good catch.

src/plugins/theme/fontFamily.js
sherakama: I don't think it hurts to have the fallback and that makes sense with what you are saying on the expectation of the project importing the fonts.

yvonnetangsu:
> GTG. I've added a `minor` tag to this PR as it does require the downstream consumer to make change.

Thanks @sherakama - totally agree about the minor tag 😄

PR #919, opened 2023-07-11
GIVCAMP-151 | Spacing options and other utilities updates

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Extend TW gap options instead of override, which allows it to have all our spacing values as options
- Add more options in theme spacing for convenience (this allows utilities such as padding, margin, width, gap etc to use these values instead of having to go with the custom [])
- Support for nested centered containers
- Add additional color variants to the Identity Guide colors that are frequently used in our projects - they are used for Decanter form elements or for accessibility contrast over existing colors.
- Add new font-basic.css file which exclude the Roboto Slab and Roboto Mono font imports for projects that don't need those fonts and only need the basic sans, serif and Stanford fonts.
- Misc fixups

# Needed By (Date)
- Not urgent but sooner is always nice

# Steps to Test

1. Look at code
2. Look at things called out
3. Look at preview site and see if things look weird anywhere

- GIVCAMP-151


README.md
```diff
@@ -1,5 +1,5 @@
 # [Decanter](https://github.com/SU-SWS/decanter)
-##### Version: 7.0.0-dev
+##### Version: 7.1.1
```
sherakama, line 2: We never keep this up to date. I've been removing it from our readmes or replacing with just the major version. The specific version in package.json is good enough.

yvonnetangsu: God idea - I 100% agree 👍🏼 😄

README.md
```diff
@@ -19,8 +19,10 @@ For Decanter v7, instead of using SASS/SCSS with the BEM naming convention, we u
 We have removed all assets from the repo in Decanter v7. Instead, we are using remote third party resources for our fonts and icons.
 
 ### Fonts
-- For Source Sans Pro, Source Serif Pro, Roboto Slab, Roboto Mono - we include them using the `@import` method from [Google Fonts](https://fonts.google.com/).
-- For the Stanford ligature font that we use for the logo, we link them from the [University Communications media CDN](https://www-media.stanford.edu/assets/fonts/stanford.woff).
+- We do not include any font assets with Decanter v7. If you want to use the Decanter fonts in your own projects, it is recommended that you use the font loading method that is optimized for your framework and only import the fonts that you need.
```
sherakama, line 22: We can also encourage our clients to import our fonts.css file in their CSS file as we do in alumni. 

https://github.com/SU-SWS/saa_alumni/blob/dev/src/styles/global.css#L1

This way, when we change, they change.

static/index.html
```diff
@@ -1295,6 +1295,25 @@ <h3 class="mb-0">2 x 1 Box</h3>
...
+
+    <div class="box">
+      <div class="box">
+        <div class="box bg-sky-light">
+          <div class="rs-p-6">
+            Mike's approach for nested cc
+          </div>
+        </div>
+      </div>
+    </div>
```
sherakama, line 1316: Thanks for adding the differences.

src/plugins/components/layout/centered-container.js
```diff
@@ -29,6 +29,10 @@ module.exports = function () {
           paddingLeft: 'calc((100% - 1500px)/2)',
           paddingRight: 'calc((100% - 1500px)/2)',
         },
+        '.cc': {
+          paddingLeft: 0,
+          paddingRight: 0,
+        },
```
yvonnetangsu, line 35: I added both Mike's approach and mine for now so you can see a comparison. This is my approach to support nested centered-container and I think it works as expected if I understand the intent. Also, I think there's some issues with support for browsers that have scroll bar by using vw unit (not obviously on Mac since scrollbar is turned off by default). We can talk more tomorrow @sherakama 

Screenshot 2023-07-11 at 4 51 27 PM

static/index.html
```diff
@@ -1295,6 +1295,25 @@ <h3 class="mb-0">2 x 1 Box</h3>
...
+
+    <div class="box">
+      <div class="box">
+        <div class="box bg-sky-light">
+          <div class="rs-p-6">
+            Mike's approach for nested cc
```
sherakama, line 1312:
```suggestion
            Mike's approach for nested containers
```

static/index.html
```diff
@@ -1295,6 +1295,25 @@ <h3 class="mb-0">2 x 1 Box</h3>
       </div>
     </section>
 
+    <div class="cc">
+      <div class="cc">
+        <div class="cc bg-poppy-light rs-mb-2">
+          <div class="rs-p-6">
+            Yvonne's approach for nested cc
```
sherakama, line 1302:
```suggestion
            Yvonne's approach for nested containers
```

static/index.html
yvonnetangsu: The original one doesn't work this way, but the newly added code below included this use case.

static/index.html
yvonnetangsu: Oh this is for internal conversation and preview only - I fully intend to clean this up @sherakama 😂

README.md
yvonnetangsu: This makes sense - I might update this to split the import into 2 files - one for essential sans and serif, the other also include the slab and monospace fonts 🤔

fonts-basic.css
```diff
@@ -0,0 +1,10 @@
+@charset "UTF-8";
+
+/** Source Sans 3, Source Serif 4 from Google Fonts */
+@import url('https://fonts.googleapis.com/css2?family=Source+Sans+3:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&family=Source+Serif+4:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap');
+@font-face {
+  font-family: "Stanford";
+  src: url("https://www-media.stanford.edu/assets/fonts/stanford.woff") format("woff"), url("https://www-media.stanford.edu/assets/fonts/stanford.ttf") format("truetype");
+  font-weight: 300;
+  font-display: swap;
+}
```
yvonnetangsu, line 10: Added fonts-basic.css which excludes the slab and mono fonts so we can be less bloated

README.md
yvonnetangsu: This has been updated to 
`##### Version: 7`

src/plugins/components/form/buttons.js
```diff
@@ -10,7 +10,7 @@ module.exports = function () {
         cursor: 'pointer',
         display: 'inline-block',
         border: 'none',
-        fontWeight: config('theme.fontWeight.regular'),
+        fontWeight: config('theme.fontWeight.normal'),
```
yvonnetangsu, line 13: I'm replacing "regular" with "normal" everywhere when referencing the fontWeight in the theme this way. "normal" is the TW default, we added an alias "regular" so people who came from Decanter v6 (that calls font-weight: 400 "regular") can feel more at home. I'd rather go with the TW default whenever possible since we might get rid of the alias in the future.

src/plugins/components/form/form-elements.js
```diff
@@ -33,19 +33,19 @@ module.exports = function () {
       '.textarea': {
         height: '16rem',
         display: 'block',
-        borderColor: theme('color.foggy-dark'),
+        borderColor: theme('colors.black.20'),
```
yvonnetangsu, line 36: Kerri mentioned that the decanter form elements needed to be updated to match My Account. So I'm having it match the My Account styles.

src/plugins/theme/spacing.js
```diff
@@ -47,43 +50,91 @@ module.exports = function () {
...
     40: '4rem',
+    41: '4.1rem',
+    42: '4.2rem',
+    43: '4.3rem',
+    44: '4.4rem',
     45: '4.5rem',
+    46: '4.6rem',
+    47: '4.7rem',
     48: '4.8rem',
+    49: '4.9rem',
```
yvonnetangsu, line 71: As much as we'd like to stick to the modular spacers, we very frequently need these extra values (esp below 50px). Sometimes to compensate for optical effects, or for finetuning positions of elements.

static/index.html
```diff
@@ -12,18 +12,14 @@
...
   </div>
-  <div role="banner" class="rs-py-6 bg-gradient-to-b from-black to-plum text-white">
-    <div class="centered-container">
-      <h1 class="splash-text">Decanter 7 With Tailwind Test Page.</h1>
-      <p class="intro-text">This is a test page filled with common HTML elements to be used to provide visual
-        feedback whilst building CSS systems and frameworks.</p>
-    </div>
+  <div role="banner" class="cc rs-py-6 bg-gradient-to-b from-black to-plum text-white">
+    <h1 class="splash-text">Decanter 7 With Tailwind Test Page.</h1>
+    <p class="intro-text">This is a test page filled with common HTML elements to be used to provide visual
```
yvonnetangsu, line 21: I'm just removing extra divs that are not needed. Putting the `cc` class in the outermost container does the same thing.

static/index.html
```diff
@@ -779,20 +775,20 @@ <h3 class="type-5 text-lagunita-light">Stone</h3>
...
-          <article id="color-palette-foggy" class="rs-mb-5">
+          <article id="color-palette-fog" class="rs-mb-5">
             <header>
-              <h3 class="type-5 text-lagunita-light">Foggy</h3>
+              <h3 class="type-5 text-lagunita-light">Fog</h3>
             </header>
             <div class="grid md:grid-cols-2 lg:grid-cols-3 gap-xl md:gap-lg 2xl:gap-2xl">
               <div
-                class="rs-py-4 rs-px-1 flex justify-center content-center font-semibold border border-black-10 text-black bg-foggy">
+                class="rs-py-4 rs-px-1 flex justify-center content-center font-semibold border border-black-10 text-black bg-fog">
```
yvonnetangsu, line 784: For a while we've added the alias `foggy` to the color `fog` since in the original color palette (decanter v5 or before), fog was a different color compared to our current fog. I think we're ok now. I'm leaving the `foggy` alias in, but we're updating our code to use the OG `fog` whenever possible.

yvonnetangsu:
@sherakama Just added a bunch of comments and write up the PR summary. Let's talk more in our 1:1 this afternoon 😄

yvonnetangsu:
@rebeccahongsf Tagging you for a FYI since you're a heavy user as well 😄

src/plugins/components/layout/centered-container.js
```diff
@@ -29,6 +29,10 @@ module.exports = function () {
           paddingLeft: 'calc((100% - 1500px)/2)',
           paddingRight: 'calc((100% - 1500px)/2)',
         },
+        '.centered-container, .cc': {
+          paddingLeft: 0,
+          paddingRight: 0,
+        },
```
sherakama, line 35: Trixy 👍

sherakama (approved):
GTG

yvonnetangsu:
@sherakama I'll talk to you more in 1:1 - we might not need the extra .box utility if .cc now does what .box do also.

src/plugins/components/typography/wysiwyg.js
```diff
@@ -13,7 +13,6 @@ module.exports = function () {
           marginBottom: '0',
         },
         'p, ol, ul, dl': {
-          color: theme('color.black.true'),
```
yvonnetangsu, line 16: I don't feel like we need to dictate what text color people use in a WYSIWYG. Original I picked the true black (#000) thinking it would improve a11y since the contrast is higher, but Kerri mentioned that it's not always the case. Default for HTML document is black text anyhow so I think it's ok for us to remove this.

PR #950, opened 2025-09-03
v8 | DS-1156 | v8 - upgrade to TW v4

yvonnetangsu:
# READY FOR REVIEW

# Summary

Rewrites Decanter from a Tailwind v3 JavaScript preset into a plain-CSS Tailwind v4
preset. `src/css/` is now the product: no JS config, no build step, no compiled
artifact in the package.

## Dependency and toolchain changes

**`tailwindcss` moved from a direct dependency to a peer dependency.**

| | main (v7) | v8 |
|---|---|---|
| `dependencies` | `tailwindcss ^3.4.17`, `@tailwindcss/aspect-ratio ^0.4.2`, `@tailwindcss/forms ^0.5.11` | `@tailwindcss/forms ^0.5.11` |
| `peerDependencies` | — | `tailwindcss ^4.1.16` |
| `devDependencies` | `postcss`, `autoprefixer`, `eslint`, `eslint-config-airbnb`, `auto-changelog`, `serve` | `tailwindcss ^4.3.3`, `@tailwindcss/cli ^4.3.3`, `concurrently`, `auto-changelog`, `serve` |
| Node engine | `>=18` | `>=20` |

- On v7, installing `decanter` pulled its own copy of Tailwind v3 in. A project on a
  different Tailwind version got a second, nested install, and the version Decanter's
  preset was written against was not necessarily the one compiling the project's CSS.
  As a peer dependency, the consumer owns the single Tailwind install and any version
  mismatch surfaces at install time instead of as duplicated or diverging output.
- Tailwind also moves into `devDependencies` here so the demo build in this repo still
  runs. Note the peer floor (`^4.1.16`) is lower than what we actually build against
  (`^4.3.3`).
- `@tailwindcss/aspect-ratio` is dropped, which restores core `aspect-*`
  (`aspect-video`, `aspect-square`, `aspect-auto`) and removes `aspect-w-*` /
  `aspect-h-*` / `aspect-none`. Bare fractions like `aspect-16/9` now work.
- `@tailwindcss/forms` stays a real dependency, because `src/css/forms.css` loads it with
  `@plugin '@tailwindcss/forms'` and Tailwind's engine resolves that from Decanter's own
  package tree at build time — so `decanter/forms` works with no extra install step.
  It is installed as a transitive dependency whether or not you use it, but it is
  build-time JS that never reaches the browser, and **nothing it generates enters your
  CSS unless you `@import 'decanter/forms'`**. Skip that import and your output contains
  neither the form classes nor the `@tailwindcss/forms` global element reset — on v7
  both were unconditional, since the preset registered the plugin for every project.

**PostCSS and autoprefixer are gone.** `postcss.config.js` (which registered the
`tailwindcss` and `autoprefixer` plugins) is deleted, along with both devDependencies.
Tailwind v4 handles vendor prefixing itself via Lightning CSS, and this repo now builds
with `@tailwindcss/cli` directly:

`tailwindcss -i dev/dev.css -o static/css/decanter.css`

Decanter imposes no PostCSS requirement on consumers — any v4 integration works
(`@tailwindcss/cli`, `@tailwindcss/vite`, `@tailwindcss/postcss`). Projects that do use
PostCSS should note the v4 migration: the plugin is `@tailwindcss/postcss`, not
`tailwindcss` + `autoprefixer` in `postcss.config.js`.

## Colors
**Renamed** (identical values — `foggy` was a v6 name kept only to avoid a class
conflict that no longer exists):
| v7 | v8 | Value |
|---|---|---|
| `foggy` | `fog` | `#DAD7CB` |
| `foggy-light` | `fog-light` | `#F4F4F4` |
| `foggy-dark` | `fog-dark` | `#B6B1A9` |

**Removed** — third-party brand colors, never part of the Stanford Identity palette and
stale as platforms rebrand (`twitter` predates X). No replacement; define the ones you
need in your own `@theme`:

| Removed | Value |
|---|---|
| `facebook` | `#4267B2` |
| `twitter` | `#1DA1F2` |
| `instagram` | `#E1306C` |
| `linkedin` | `#0077B5` |
| `youtube` | `#FF0000` |

- Tailwind's own palette stays available alongside Decanter's in both versions — v7's
  `extend.colors` deep-merged, so despite the "replaces the TW Sky/Stone" comments in
  `colors.js`, `sky-500` and `stone-500` kept working there and still do in v8.

## New in v8
- **Responsive gap utilities:** `rs-gap-*`, `rs-gap-x-*`, `rs-gap-y-*` (steps 0–10),
  filling the gap in the `rs-*` family — v7 had responsive padding and margin only, so
  grid and flex gutters had to be hand-written with breakpoint modifiers.
- **`type-10`**, extending the modular type scale that stopped at `type-9` in v7.
- **`hocus-within` variant**, which also gives `group-hocus-within`.
- **Overridable theme variables.** Values that were locked inside the JS config are now
  `@theme` custom properties — including `--screen-margin-*`, which projects can
  redefine to change `centered-container` gutters without forking anything.
- **New entry points:** `decanter/minimal` (theme without element-level base opinions)
  and `decanter/colors` (palette only, works on stock Tailwind).

## Other consumer-facing changes
- **The JS preset is gone.** `tailwind.config.js` / `.mjs` / `.d.ts` and `tw.config.js`
  are deleted, along with all 30+ files under `src/plugins/`. Consumers drop
  `presets: [require('decanter')]` and `@import 'decanter'` instead. Decanter's main
  entries include `@import 'tailwindcss'` internally — importing Tailwind again
  duplicates preflight.
- **Forms are opt-in.** `.input`, `.select`, `.textarea`, `.checkbox`, `.radio`,
  `.label`, `.legend`, `.fieldset` and the forms reset moved to `decanter/forms`. Sites
  without forms no longer ship the global element reset.
- **Fonts unbundled.** `fonts.css` / `fonts-basic.css` removed; Decanter ships the
  `font-sans` / `font-serif` / `font-stanford` utilities and documents how to load the
  families.
- **Other removals and renames** — full list in [UPGRADE.md](UPGRADE.md): `font-slab`
  removed, `font-mono` no longer overridden, `text-m0`–`text-m9` removed (use `type-*`),
  `.embed-container` → `aspect-video` / `aspect-16/9`, `.credits` removed, negative
  `rs-*` steps removed, `break-words` → `wrap-anywhere`, `font-regular` →
  `font-normal`, `link-regular` → `link-normal`, `text-shadow-*` →
  `text-shadow-legacy-*`, `children:` → Tailwind's `*:`. All were deprecated in a 7.5.x
  release, so a project that migrated on v7 upgrades with no changes.

## Repo changes
- ESLint, its config, and the lint workflow removed — there is no JS left to lint.
- `npm run dev` uses `concurrently` for watch + serve.
- Docs restructured from per-plugin files into topic guides (getting-started, colors,
  typography, spacing-and-layout, components, forms, utilities, variants).
- `static/index.html` rebuilt against the new class surface with a fresh design

# Needed By (Date)
- Soon

# Urgency
- 5

# Steps to Test

1. Look at new preview page
https://deploy-preview-950--decanter-v7.netlify.app/
2. Check that the TW custom Decanter utilities are all working
3. Look at the UPGRADE.md and all other new docs
4. Take a look at this Open Minds test branch that uses the v8 branch commit b9365c1
https://github.com/SU-SWS/ood-stanford-tour/pull/347

- DS-1156


src/css/utilities.css
```diff
@@ -0,0 +1,288 @@
...
+@utility link-bold {
+  a {
+    font-weight: var(--font-weight-bold);
+  }
+}
+
+/**
+  Responsive spacing - padding
+*/
+@utility rs-p-* {
```
sherakama, line 4: Interesting. So this is how they do their dynamic property values. Simple enough.

src/css/custom-variants.css
```diff
@@ -0,0 +1,27 @@
+/**
+  Custom variants
```
sherakama, line 2: Neat.

src/css/base.css
sherakama, line 1: Does TW4 ship with a css reset still?

src/css/components.css
```diff
@@ -0,0 +1,673 @@
...
+
+    @media (min-width: 1500px) {
+      font-size: 2.3rem;
+    }
+  }
+
+  .wysiwyg {
+    > *:first-child {
+      margin-top: -0.25em;
+    }
```
sherakama: I guess one nice thing about moving to a CSS based config is that you can write natural css again.

src/css/dev.css
```diff
@@ -0,0 +1,10 @@
+/** Source Sans 3, Source Serif 4, Roboto Slab, Roboto Mono (regular only) */
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto+Slab:wght@300;400;700&family=Source+Sans+3:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&family=Source+Serif+4:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap');
+
+@import 'tailwindcss';
+@source not '../../docs';
```
sherakama: Huh. Funny syntax. [docs](https://tailwindcss.com/docs/detecting-classes-in-source-files#ignoring-specific-paths)

src/css/theme.css
```diff
@@ -0,0 +1,226 @@
+@theme {
```
sherakama: How you feeling about having all of these values pulled from a Figma file one day?

sherakama: Any thoughts on having a core and semantic layer?

src/css/dev.css
yvonnetangsu: Some of these things were hard to find 😬

src/css/components.css
```diff
@@ -0,0 +1,673 @@
...
+
+    .centered-container,
+    .cc {
+      padding-left: 0;
+      padding-right: 0;
+    }
+  }
+
+  /* Responsive grid gap */
+  .grid-gap {
```
sherakama: I could argue this is more of a utility than a component. But I won't die on that hill.

src/css/components.css
```diff
@@ -0,0 +1,673 @@
...
+    text-shadow: rgba(0, 0, 0, 40%) 0 0 6px, rgba(0, 0, 0, 60%) 0 0 2px;
+  }
+
+  .text-shadow-md {
+    text-shadow: rgba(0, 0, 0, 40%) 0 0 8px, rgba(0, 0, 0, 60%) 0 0 3px;
+  }
+
+  .text-shadow-lg {
+    text-shadow: rgba(0, 0, 0, 30%) 0 0 12px;
+  }
```
sherakama: Utilities imho.

src/css/utilities.css
yvonnetangsu: Yeah I'm glad they have these - compensate for no js 👍🏼

src/css/components.css
yvonnetangsu: I'm going to think about what to do with these because TW has core support for text shadows now

src/css/components.css
yvonnetangsu: Yeah the reason I put it in component is that you can override it with the core utilities, so for example, if using `grid-gap gap-y-10`, the gap-y-10 will override the vertical gap in grid-gap

src/css/theme.css
yvonnetangsu: Still playing around with things 😄  Going to think about everything some more

src/css/components.css
yvonnetangsu: But yeah this one is a utility because it doesn't need to be overridden by anything else 👍🏼

sherakama:
With presets gone, is it now just `@import decanter` in the project stylesheet?

src/css/utilities.css
```diff
@@ -0,0 +1,288 @@
...
+  row-gap: --value(--responsive-spacing-xs-*);
+
+  @media (min-width: 768px) {
+    row-gap: --value(--responsive-spacing-md-*);
+  }
+
+  @media (min-width: 1500px) {
+    row-gap: --value(--responsive-spacing-2xl-*);
+  }
+}
```
yvonnetangsu, line 224: Justin asked for this so adding it

yvonnetangsu:
> With presets gone, is it now just `@import decanter` in the project stylesheet?

Yup, from what I understand, it's basically just importing what you need 😬

src/css/base.css
yvonnetangsu: I plan on thinking a bit more about this as well 🤔

src/css/utilities.css
```diff
@@ -0,0 +1,288 @@
...
+
+@utility text-* {
+  font-size: calc(--value(integer) * var(--font-size));
+}
```
yvonnetangsu: Font size utility - it supports any integer values, so `text-100` would be `font-size: 10rem`. It's an extension so all our custom font size classes, eg, `text-1em`, still works.

src/css/components/all-no-forms.css
sherakama, line 1: My $0.02, kill this file. 

Allow someone to easily import everything or individual items. If forms is an issue, perhaps look at breaking it up and into individual components too.

src/css/components/typography-styles.css
```diff
@@ -0,0 +1,162 @@
+/**
+  Decanter typography styles
+*/
+
+@layer components {
+  .types
```
sherakama:
```suggestion
  .types,
```

sherakama: Oh wait, this is a descendant selector. Why do we need `.types .splash-text` and `.splash-text`?

yvonnetangsu: Oh I haven't cleaned up a lot of the stuff - some of it I had AI move things around 😂 I'll go back and see if I can rewrite the modular type stuff in a cleaner way too.
What are you doing here by the way @sherakama ? 👀

src/css/components/all-no-forms.css
yvonnetangsu: Sure I'll kill it 🔪

src/css/components/typography-styles.css
yvonnetangsu: I want to rethink some of this typography styles btw

src/css/utilities.css
iamrentman: Oh hey! Neat. Thank you!

src/css/theme.css
yvonnetangsu: > Any thoughts on having a core and semantic layer?

I'm thinking that Decanter core wouldn't have a semantic layer, because other than maybe `--color-brand` would be universal, we can't really predict what other projects will use for background or default border or subtle (can be fog-light, can be black-10 or a custom color like for homesite) 🤔 Perhaps the semantic layer will be defined in individual projects @sherakama and tagging @iamrentman

src/css/base.css
yvonnetangsu: TW4 by default does come with preflight/reset styles:
https://tailwindcss.com/docs/preflight

It is possible to exclude it, perhaps I'll set up an import option without the preflight.

src/css/theme.css
sherakama: That sounds reasonable to me. 

One thought as you use this and learn more, perhaps create an optional semantic layer in which other teams could opt-in to. eg:

```
@import "decanter";
@import "decanter/semantic-vars";
```

yvonnetangsu: @sherakama I could for sure, perhaps UComm @iamrentman could work with us to define a set of semantic variables?

src/css/base.css
sherakama: Well, if the expectation is that this package is always imported into another tailwind projet, perhaps we never include it here? 

That is only important, however, if the preflight is being doubled up on the export. 

If it is coming out the end with just one set of resets, then all is good.

src/css/theme.css
iamrentman: Thanks for calling this out. I was wondering the same thing as I browse this and better understand Tailwind v4. I think you're right, @yvonnetangsu – semantic layer is in the project utilizing these primitive values.

src/css/base.css
yvonnetangsu: Yeah good point - I'll double check 👍🏼

src/css/theme.css
yvonnetangsu: Hi @iamrentman 👋🏼 If you feel like we could add a "recommended" set of semantic variables, we can certainly add that. There is no rush, we can think about this some more - I'm also thinking about things.

src/css/theme/responsive-spacing.css
```diff
@@ -0,0 +1,41 @@
+@theme {
+  --responsive-spacing-xs-neg2: 0.8rem;
+  --responsive-spacing-md-neg2: 0.9rem;
+  --responsive-spacing-2xl-neg2: 1rem;
+  --responsive-spacing-xs-neg1: 1.1rem;
+  --responsive-spacing-md-neg1: 1.2rem;
+  --responsive-spacing-2xl-neg1: 1.3rem;
```
yvonnetangsu, line 7: I'm thinking about deprecating these negative values. I'll add doc in the upgrade.md, e.g., 
`rs-p-neg1 => p-11 md:p-12 2xl:p-13`
What do you think? @iamrentman @sherakama

sherakama: Yeah, I am find with that as I feel they break convention. 

Deprecate them in V7, and remove in V8.

yvonnetangsu: Sounds good @sherakama - I'll start a list of things to be deprecated in decanter v7

src/css/utilities/fluid-type.css
```diff
@@ -0,0 +1,59 @@
...
+  * XS: 1.8rem, MD: 2.1rem, 2XL: 2.3rem
+  */
+
+@utility fluid-type-0 {
+  font-size: clamp(1.8rem, 0.44vw + 1.64rem, 2.3rem);
+}
+
+@utility fluid-type-1 {
+  font-size: clamp(2.1rem, 0.7vw + 1.85rem, 2.9rem);
+}
```
yvonnetangsu, line 23: Add fluid type as utilities instead of components. Apparently in TW v4, things like lg:fluid-type-4 (when you use a modifier) stops working if fluid-type is a component.

yvonnetangsu:
Testing latest commit using Open Minds - seems to work fine
https://deploy-preview-347--stanford-tour.netlify.app/

package.json
```diff
@@ -1,53 +1,41 @@
 {
   "name": "decanter",
-  "description": "Design System Tailwind CSS Preset.",
+  "description": "Stanford University's design system as a CSS-first Tailwind CSS v4 preset: color palette, typography, spacing, components, and utilities.",
   "author": "Stanford Web Services",
-  "version": "7.5.3",
+  "version": "8.0.0-alpha.1",
```
yvonnetangsu, line 5: This was updated to 8.0.0-alpha.1 so when we do autorelease, it will be bumped to alpha.2

yvonnetangsu: @sherakama We might even be ready for a beta release - let me know what you think 👀

sherakama: shaq-cruise

Ship it.

yvonnetangsu: Ok I'll change the tag to beta 🤠

yvonnetangsu:
`main` has been merged into `v7` (it's the same as release 7.5.3 now)

yvonnetangsu:
Ok ready to merge this beta version for real. The v8 code here also replace the `main` branch, but I think it's time.

PR #949, opened 2025-07-28
ES/Common and documentation.

sherakama:
# READY FOR REVIEW

# Summary
- Exports two sets of files (ES & Common)
- Absolute path includes to support compilers such as turbopack
- Attempts to maintain backward compatibility
- Adds documentation

# Needed By (Date)
- Whenever you can

# Urgency
- None

# Steps to Test

1. Add this branch to a few older sites and ensure it still compiles with no additional modifications
2. Add this branch to a newer NextJS site (emergency maybe?) 
3. Update the tailwind.config.(t/mj/j)s to an ES module and change any imports/requires necessary
4. Add `--turbopack` to the `npm run dev/build` commands
5. Ensure they run with no errors

## Part 2

1. Read through (some) of the documentation

PR #959, opened 2026-08-17
v7 | Fix npm publish with trusted publishing; update GitHub Actions to v7; remove unused packages

yvonnetangsu:
# READY FOR REVIEW

# Summary

Fixes the npm publish failure that left 7.5.0 tagged on GitHub but never published, brings the GitHub Actions off deprecated runtimes, repairs `npm run dev`, and corrects outdated docs.

No functional change for consumers — nothing under `src/`, `tailwind.config.*`, `tw.config.js`, or the font CSS is touched. The only shipped changes are documentation and inert `package.json` fields.

## Publishing

- Switch to npm **trusted publishing** (OIDC) instead of a stored `NPM_TOKEN`. The 7.5.0 publish failed with an opaque `status of 1`, which was a dead token — npm revoked classic tokens and caps granular ones at 90 days. Requires `id-token: write`, `registry-url`, and npm ≥ 11.5.1.
- Call `npm publish` directly rather than through `JS-DevTools/npm-publish`, which was swallowing npm's actual error.
- Derive the dist-tag from the version so prereleases publish under their preid and never clobber `latest`.
- Replace the removed `set-output` command with `$GITHUB_OUTPUT`, using the heredoc form. `$GITHUB_OUTPUT` does no percent-decoding, so the old escaping would have written literal `%0A` into every release note.

> **Note:** the trusted publisher must be registered on npmjs.com (decanter → Settings → Trusted Publisher → this repo + `autorelease.yml`) or the publish fails the same way.

## GitHub Actions

- `actions/checkout` and `actions/setup-node` → `v7` across all three workflows. The old `v2`/`v3` pins run on Node 16, and `v4` declares Node 20, which is **removed from runners on 2026-09-16**.
- `build.yml` and `lint.yml` now read the Node version from `.nvmrc` instead of pinning `18.x` (EOL April 2025).
- Fix `build.yml`'s artifact check: `git diff --exit-code` fails the step immediately under `set -e`, so the `if [ $? -ne 0 ]` branch below it never ran and failures printed no explanation.

## Dev environment

- **`npm run dev` has been broken since 7.5.0.** It backgrounds the watcher, and a backgrounded job in a non-interactive shell gets stdin from `/dev/null`. Tailwind's `--watch` treats stdin EOF as a shutdown signal, so it exited before completing a single build — developers got a working server quietly serving stale CSS. `--watch=always` is Tailwind's flag for exactly this case.
- Replace `light-server` (last published 2019, 10 of the repo's advisories via `gaze → globule → minimatch`) with `serve` 14.2.6 (0 advisories). Same port 4000, installed as a devDependency so `npm ci` stays deterministic.
- Remove `typescript`, `@types/node` and `eslint-config-standard`. The first two arrived with #949 alongside `tailwind.config.d.ts`, but no `tsconfig.json` or `tsc` script ever landed. `eslint-config-standard` was never loaded — `.eslintrc` extends only `airbnb`.
- `auto-changelog` 2.4.0 → 2.6.0 (lockfile-only; `^2.4.0` already permitted it), pulling `handlebars` 4.7.9 over the vulnerable pin.

Advisories **25 → 14**, critical **1 → 0**. All remaining ones are dev-only.

`eslint` deliberately stays on 7 — reaching 9 means migrating to flat config, and v8 already deletes `.eslintrc` outright.

## Documentation

- **`README.md`**: correct the font filenames (`fonts.css` / `fonts-basic.css` — the README said `font.css` / `font-basic.css`, so anyone following it got a module-not-found); point at `docs/installation.md` (`docs/usage.md` doesn't exist); replace the 404'd accessibility policy link; add an Installation section; link contributors to `docs/development.md` and `CONTRIBUTING.md`.
- **`docs/css-files.md`**: drop `import 'decanter/dist/decanter.css'` — there is no `dist/` and it isn't in the `files` allowlist, so it could never resolve.
- **`docs/development.md`**: replace the non-existent `node test-commands` check; add a Local Development section (the guide had no way to run anything); document that there's no live reload, and that the watcher can't see `src/plugins/` changes because `tw.config.js` loads plugins via an injected `requireFn(...)` that Tailwind's static config scan can't follow.
- **`static/index.html`**: update four stale Stanford footer URLs. Remaining churn is an editor reflow — verified no semantic change (normalized to one tag per line, both versions are 2148 lines and differ only in those URLs).

## Verification

- `npm ci`, `npm run lint`, `npm run build` all pass from a clean `node_modules`; build regenerates no tracked files.
- `npm run dev`: server responds on :4000, `src/css/index.css` edits rebuild and are served over HTTP, Ctrl-C leaves no orphaned processes.
- Consumer typecheck against a packed tarball in a fresh project (`moduleResolution: nodenext`, `strict`) resolves `decanter` → `tailwind.config.d.ts` and rejects invalid assignments, confirming removing `typescript` doesn't affect TS consumers.
- `dependencies` untouched; tarball still 95 files.
- All README links resolve; all external URLs return 200.

## Release label

`patch` → **7.5.1**. Docs-only from a consumer's perspective, and the `font.css` filename error is currently live on the npmjs.com package page.

# Needed By (Date)
- Soon

# Urgency
- 5

# Steps to Test

1. Quick check that this is still working
https://deploy-preview-959--decanter-v7.netlify.app/
2. Look at code and updated doc

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,12 +19,12 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
 
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
-          node-version: '18.x'
+          node-version-file: '.nvmrc'
 
       - name: Install dependencies
         run: npm ci
--- a/README.md
+++ b/README.md
@@ -18,6 +18,16 @@ This repository includes extensive documentation to help you understand and use 
 1. **[README.md](README.md)** - Overview of Decanter, installation instructions, and quick start guide.
 2. **[docs/README.md](docs/README.md)** - Main documentation index with links to detailed guides.
 
+If you want to work on Decanter itself rather than use it in a project:
+- **[docs/development.md](docs/development.md)** - Architecture, the shared config factory, and how to run the package locally.
+- **[CONTRIBUTING.md](CONTRIBUTING.md)** - Issue reporting, feature requests, and pull request guidelines.
+
+### Installation
+
+```bash
+npm install decanter
+```
+
 ### Usage
 When using Decanter, the package will automatically resolve to the correct module format:
 
@@ -28,7 +38,7 @@ module.exports = {
   // your additional config...
 }
 
-// ES Module - In your tailwind.config.mjs  
+// ES Module - In your tailwind.config.mjs
 import decanter from 'decanter';
 export default {
   presets: [decanter],
@@ -38,10 +48,10 @@ export default {
 
 The package exports both CommonJS (`tailwind.config.js`) and ES Module (`tailwind.config.mjs`) versions to ensure compatibility with different bundlers and configurations.
 
-For detailed usage examples and configuration options, see [docs/usage.md](docs/usage.md).
+For detailed usage examples and configuration options, see [docs/installation.md](docs/installation.md).
 
 ## CSS Class Index
-Comprehensive list of all custom CSS classes provided by Decanter.  
+Comprehensive list of all custom CSS classes provided by Decanter.
 See [docs/index.md](docs/index.md)
 
 ## Assets
@@ -49,7 +59,7 @@ We have removed all assets from the repo in Decanter v7. Instead, we are using r
 
 ### Fonts
 - We do not include any font assets with Decanter v7. If you want to use the Decanter fonts in your own projects, it is recommended that you use the font loading method that is optimized for your framework and only import the fonts that you need.
-- We provided two font css files that you can import into your own project so you get the latest update from us. `font.css` includes all the fonts that are referenced in the Decanter design system (sans-serif, serif, slab, monospace, Stanford ligature font for the logo). `font-basic.css` includes only the essential sans-serif, serif and Stanford ligature fonts.
+- We provided two font css files that you can import into your own project so you get the latest update from us. `fonts.css` includes all the fonts that are referenced in the Decanter design system (sans-serif, serif, slab, monospace, Stanford ligature font for the logo). `fonts-basic.css` includes only the essential sans-serif, serif and Stanford ligature fonts.
 - For Source Sans 3, Source Serif 4, Roboto Slab, Roboto Mono - we include them using the `@import` method from [Google Fonts](https://fonts.google.com/).
 - The Stanford ligature font that we use for the logo is linked from the [University Communications media CDN](https://www-media.stanford.edu/assets/fonts/stanford.woff).
 
@@ -60,4 +70,4 @@ We have removed all assets from the repo in Decanter v7. Instead, we are using r
 ## Accessibility
 [![WCAG Conformance 2.0 AA Badge](https://www.w3.org/WAI/wcag2AA-blue.png)](https://www.w3.org/TR/WCAG20/)
 
-This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://ucomm.stanford.edu/policies/accessibility-policy.html](https://ucomm.stanford.edu/policies/accessibility-policy.html).
+This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://www.stanford.edu/accessibility](https://www.stanford.edu/accessibility).
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
   ],
   "scripts": {
     "build": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --content static/*.html",
-    "watch": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --watch --content static/*.html",
-    "serve": "light-server -s static -w static/*.html",
+    "watch": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --watch=always --content static/*.html",
+    "serve": "serve static -l 4000",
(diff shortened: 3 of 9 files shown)


.github/workflows/lint.yml
```diff
@@ -19,12 +19,12 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
```
yvonnetangsu, line 22: Using v7 for all of these

.github/workflows/lint.yml
```diff
@@ -19,12 +19,12 @@ jobs:
...
 
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
-          node-version: '18.x'
+          node-version-file: '.nvmrc'
```
yvonnetangsu, line 27: Have it reference the node version in .nvmrc rather than a specific version

package.json
```diff
@@ -43,8 +43,8 @@
   ],
   "scripts": {
     "build": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --content static/*.html",
-    "watch": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --watch --content static/*.html",
-    "serve": "light-server -s static -w static/*.html",
+    "watch": "tailwind -c tailwind.config.js -i src/css/index.css -o static/css/decanter.css --watch=always --content static/*.html",
+    "serve": "serve static -l 4000",
```
yvonnetangsu, line 47: light-server is super old and haven't been updated for years so just replacing it. Doesn't matter much soon anyways but want to leave it in a good state before we move to v8

README.md
```diff
@@ -38,18 +48,18 @@ export default {
 
 The package exports both CommonJS (`tailwind.config.js`) and ES Module (`tailwind.config.mjs`) versions to ensure compatibility with different bundlers and configurations.
 
-For detailed usage examples and configuration options, see [docs/usage.md](docs/usage.md).
```
yvonnetangsu, line 41: We don't actually have a usage.md so linking to installation.md which has examples too

README.md
```diff
@@ -38,18 +48,18 @@ export default {
...
 See [docs/index.md](docs/index.md)
 
 ## Assets
 We have removed all assets from the repo in Decanter v7. Instead, we are using remote third party resources for our fonts and icons.
 
 ### Fonts
 - We do not include any font assets with Decanter v7. If you want to use the Decanter fonts in your own projects, it is recommended that you use the font loading method that is optimized for your framework and only import the fonts that you need.
-- We provided two font css files that you can import into your own project so you get the latest update from us. `font.css` includes all the fonts that are referenced in the Decanter design system (sans-serif, serif, slab, monospace, Stanford ligature font for the logo). `font-basic.css` includes only the essential sans-serif, serif and Stanford ligature fonts.
+- We provided two font css files that you can import into your own project so you get the latest update from us. `fonts.css` includes all the fonts that are referenced in the Decanter design system (sans-serif, serif, slab, monospace, Stanford ligature font for the logo). `fonts-basic.css` includes only the essential sans-serif, serif and Stanford ligature fonts.
```
yvonnetangsu, line 62: missing s in file names

README.md
```diff
@@ -60,4 +70,4 @@ We have removed all assets from the repo in Decanter v7. Instead, we are using r
 ## Accessibility
 [![WCAG Conformance 2.0 AA Badge](https://www.w3.org/WAI/wcag2AA-blue.png)](https://www.w3.org/TR/WCAG20/)
 
-This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://ucomm.stanford.edu/policies/accessibility-policy.html](https://ucomm.stanford.edu/policies/accessibility-policy.html).
+This project conforms to level AA WCAG 2.0 standards as required by the university's accessibility policy. For more information on the policy please visit: [https://www.stanford.edu/accessibility](https://www.stanford.edu/accessibility).
```
yvonnetangsu, line 73: The ucomm link is no longer active so replacing

package.json
```diff
@@ -63,15 +63,12 @@
...
   "devDependencies": {
-    "@types/node": "^20.19.9",
     "auto-changelog": "^2.4.0",
     "autoprefixer": "^10.5.4",
     "eslint": "^7.32.0",
     "eslint-config-airbnb": "^18.2.1",
-    "eslint-config-standard": "^16.0.3",
-    "light-server": "^2.9.1",
     "postcss": "^8.5.26",
-    "typescript": "^5.8.3"
```
yvonnetangsu, line 74: Removed `@types/node`, `typescript`, `eslint-config-standard`. The first two arrived with #949 alongside `tailwind.config.d.ts`, but no `tsconfig.json` or `tsc` script ever landed. `eslint-config-standard` was never loaded — `.eslintrc` extends only `airbnb`.

.github/workflows/lint.yml
sherakama: I like this

package.json
sherakama: Looks like it works.

package.json
sherakama: Super. I guess we don't need typescript to include typescript types...

sherakama (approved):
Ty. GTG

package.json
yvonnetangsu: Yes I asked it to confirm that this removal won't affect consumers who use TS and it says just having the definition should be good 👍🏼

yvonnetangsu:
Ok moment of truth let's test this autorelease out 🤠

yvonnetangsu:
Works 🎉 
https://www.npmjs.com/package/decanter?activeTab=readme

sherakama:
Winner winner!

